PR #3, opened 2021-09-11
setup: created students table

smellslikekeenspirit:
Configured a postgres database locally and created the table in it through Pool

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -43,7 +43,7 @@ app.get('/', (request, response) => {
 response.json({ info: 'This is our empty homepage.'})
 })
 
-app.get('/users', getStudents)
+app.get('/students', getStudents)
 
 app.listen(port, () => {
 console.log(`App running on port ${port}.`)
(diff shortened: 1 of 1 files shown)


MatthewGalan (approved):
Instead of storing the major and college with each student, what if we had a separate table of all the majors (with their corresponding colleges & a major ID), and then just stored the major ID for each student?

I know very little about database management, though, so I'm not married to the idea.

It's looking great, though. Thanks for getting a head start on this!

smellslikekeenspirit:
@MatthewGalan I like the idea about a table of majors and corresponding colleges. Major IDs might hurt readability for the admin and they would have to keep referring to the other table right? Unless we are creating a view for the admin that includes filtering functionality, where they can perform the analytics they wanted to. We can use the 4-letter codes RIT uses for majors though if we are to implement the major ID idea.

MatthewGalan:
> Major IDs might hurt readability for the admin and they would have to keep referring to the other table right?

Admins won't be looking directly at the backend, though, right? We would return the actual major name for the frontend.

> We can use the 4-letter codes RIT uses for majors though if we are to implement the major ID idea.
Good idea!

smellslikekeenspirit:
Yeah, as long as we are creating a view for the admin sounds aok

PR #6, opened 2021-09-14
Created reusable button component

MatthewGalan:
- Created a reusable button component
- Button can be of type `ButtonType.Normal` (orange) or `ButtonType.Destructive` (red)
- Buttons can be disabled
- Added unit tests for said button component

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,7 @@
 **/node_modules
 process.env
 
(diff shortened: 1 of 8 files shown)


pjw9963 (approved):
React feels a bit foreign to me still, but I like the components and especially the tests

PR #8, opened 2021-09-15
Add text field component

MatthewGalan:
- Very basic reusable text field component.
- Tests included
- `disabled` not implemented but it would be easy and we can add it if/when we need it

image

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -1,19 +1,10 @@
 import React from "react";
-import { DestructiveButton, NormalButton } from "./components/common/Button";
+import styled from "styled-components";
+
+const StyledDiv = styled.div``;
 
 function App() {
-  return (
-    <div>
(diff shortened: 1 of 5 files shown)


smellslikekeenspirit:
looks good, can we start integrating a UI library though? The error state would have been much easier to handle if we simply used this, for instance, right? https://material-ui.com/components/text-fields/

smellslikekeenspirit (commented):
looks good, can we start integrating a UI library though? The error state would have been much easier to handle if we simply used this, for instance, right? https://material-ui.com/components/text-fields/

MatthewGalan:
Sure, we can switch to Material UI. Originally I thought it would be easier to just build our own, since we really only needed buttons and text fields, but using an existing component library would definitely make our codebase smaller.

I'll close this PR and add Material UI tomorrow.

PR #18, opened 2021-11-01
add card reader proof of concept

MatthewGalan:
- The card reader types `;` before typing the ID number
- I put a global `keydown` listener, which waits until `;` is typed, then writes down the following numbers
- Then a modal pops up showing the ID number from the card
- Then it sends it to my personal database so I can commit financial fraud later

### DIFF
--- a/client/src/common/PrettyModal.tsx
+++ b/client/src/common/PrettyModal.tsx
@@ -1,0 +1,34 @@
+import React, { ReactNode } from "react";
+import { Card, Modal } from "@mui/material";
+
+interface PrettyModalProps {
+  open: boolean;
+  onClose: () => void;
+  width?: number;
+  children: ReactNode;
+}
+
+export default function PrettyModal({
+  width = 400,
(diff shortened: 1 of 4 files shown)


pjw9963 (approved):
looks pretty good. What happens if someone pushes a ';' key by accident tho?

MatthewGalan:
> looks pretty good. What happens if someone pushes a ';' key by accident tho?

Good question. The buffer will just keep filling up with whatever keys they press. If a card is then swiped, the buffer will be cleared just before reading the number, so it shouldn't interfere.

If someone presses `;` and _then_ presses `?`, then we would indeed be receiving junk. We'll definitely need to handle this case, as well as cases where the card reader messes up.

Merging this for now, though, since this is mainly intended as a proof-of-concept for tomorrow's demo.

MatthewGalan:
[Card Swipe](https://trello.com/c/Ndlr8mkw/37-card-swipe)

PR #21, opened 2021-11-04
resolved docker issues + added machines migration

smellslikekeenspirit:
smol PR, gargantuan task really was to get docker rolling locally to be able to run migrations


server/src/server.ts
```diff
@@ -30,5 +30,5 @@ const server = new ApolloServer({
 
 const httpServer = createServer(app);
 httpServer.listen({ port: PORT }, (): void =>
-  console.log(`🚀GraphQL-Server is running on http://localhost:3000/graphql`)
+  console.log(`🚀GraphQL-Server is running on http://localhost:z${PORT}/graphql`)
```
pjw9963, line 33: Do you need the "z" in-front of the ${PORT} ?

server/src/db/migrations/20211102194503_machines_table.ts
```diff
@@ -0,0 +1,50 @@
...
+        }
+    });
+
+    knex.schema.hasTable('Machines').then(function(exists) {
+        if (!exists) {
+          return knex.schema.createTable('Machines', function(t) {
+            t.increments('id');
+            t.string('name', 50);
+            t.integer('category');
+            t.foreign('category').references('id').inTable('MachineFamilies');
```
pjw9963, line 24: I could be wrong here, but I think since you already added the column `category` you don't need to re-add the column using `t.foreign`. Instead just do `t.integer('category').references('id').inTable('MachineFamilies');`. I tried doing it this way initially too but I had issues with CASCADE not working

server/src/db/migrations/20211102194503_machines_table.ts
```diff
@@ -0,0 +1,50 @@
...
+            t.string('room', 5);
+            t.date('date_added').defaultTo(knex.fn.now());
+            t.boolean('reserved').defaultTo(false);
+            t.uuid('reserved_by');
```
pjw9963, line 28: Would it makes sense to have a separate table for reservations? This implementation would only allow the current reservation to be known, while a reservation table could have past and future reservations for each machine 

for example: 
| reservation_id | machine_id | student_id | starttime | endtime |

pjw9963 (commented):
Good start. Let me know if my comments make sense

server/src/server.ts
smellslikekeenspirit: oops, typo

server/src/db/migrations/20211102194503_machines_table.ts
smellslikekeenspirit: Yep, that worked. Thanks!!

smellslikekeenspirit:
@pjw9963 I added the `Reservations` table as well as a better way of handling promises so transaction queries are run properly. Also changed the `id` field in the `Students` table to autoincrement

pjw9963 (approved):
looks good, i'll start writing my migrations like this too so we don't get that transaction error

PR #24, opened 2021-11-16
Storefront backend

pjw9963:
Basic graphql schema definition for the storefront backend. Let me know if I forgot something or if somethings wrong

### DIFF
--- a/server/src/schemas/storeFrontSchema.ts
+++ b/server/src/schemas/storeFrontSchema.ts
@@ -2,73 +2,72 @@ import { gql } from "apollo-server-express";
 
 export const StoreFrontTypeDefs = gql`
 
-	scalar Date
+  scalar Date
 
-	type InventoryItem {
-		id: ID!;
-		image: String;
-		name: String;
-		labels: [String];
-		unit: String;
-		pluralUnit: String;
-		count: Int;
-		pricePerUnit: Float;
-	}
+  type InventoryItem {
+    id: ID!;
+    image: String;
+    name: String;
+    labels: [String];
+    unit: String;
+    pluralUnit: String;
+    count: Int;
+    pricePerUnit: Float;
+  }
 
-	input InventoryItemInput {
-		image: String;
-		name: String;
-		labels: [String];
-		unit: String;
-		pluralUnit: String;
-		count: Int;
-		pricePerUnit: Float;
-	}
+  input InventoryItemInput {
+    image: String;
+    name: String;
+    labels: [String];
+    unit: String;
+    pluralUnit: String;
+    count: Int;
+    pricePerUnit: Float;
+  }
 
-	type PurchaseOrderItem {
-		id: ID!;
-		item: InventoryItem;
-		count: Int;
-	}
+  type PurchaseOrderItem {
+    id: ID!;
+    item: InventoryItem;
+    count: Int;
+  }
 
-	type PurchaseOrder {
-		id: ID!;
-		creator: Person;
-		createDate: Date;
-		expectedDeliveryDate: Date;
-		items: [PurchaseOrderItem];
-		attachments: [String];
-	}
+  type PurchaseOrder {
+    id: ID!;
+    creator: Person;
+    createDate: Date;
+    expectedDeliveryDate: Date;
+    items: [PurchaseOrderItem];
+    attachments: [String];
+  }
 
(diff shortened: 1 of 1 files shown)


pjw9963:
[Storefront](https://trello.com/c/z99OnVpc/88-storefront)

server/src/schemas/storeFrontSchema.ts
```diff
@@ -0,0 +1,74 @@
...
+		addItemToPurchaseOrder(POId: ID!, itemId: ID!, count: Int): PurchaseOrder
+		removeItemToPurchaseOrder(POId: ID!, itemId: ID!, count: Int): PurchaseOrder
+		updateItemAmountInPurchaseOrder(POId: ID!, POItemId: ID!, count: Int!): PurchaseOrder
+		addAttachmentsToPurchaseOrder(POId: ID!, attachments: [String]): PurchaseOrder
+		removeAttachmentFromPurchaseOrder(POId: ID!, attachment: String): PurchaseOrder
+		addItemToInventory(item: InventoryItemInput): InventoryItem
+		addItemAmount(itemId: ID!, count: Int!): InventoryItem
+		removeItemAmount(itemId: ID!, count: Int!): InventoryItem
+	}
+`;
```
MatthewGalan, line 74: Indented with tabs and no newline at end of file.

server/src/schemas/storeFrontSchema.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export const StoreFrontTypeDefs = gql`
+
+	scalar Date
+
+	type InventoryItem {
+		id: ID!;
+		image: String;
+		name: String;
+		labels: [String];
```
MatthewGalan, line 11: Should we have a `Label` type instead of just strings? Just asking - I could be convinced either way.

server/src/schemas/storeFrontSchema.ts
```diff
@@ -0,0 +1,74 @@
...
+		creator: Person;
+		createDate: Date;
+		expectedDeliveryDate: Date;
+		items: [PurchaseOrderItem];
+		attachments: [String];
+	}
+
+	input PurchaseOrderInput {
+		creator: Person;
+		createDate: Date;
```
MatthewGalan, line 45: `createDate` will be generated by the system, not inputted by the user, no?

server/src/schemas/storeFrontSchema.ts
```diff
@@ -0,0 +1,74 @@
...
+		expectedDeliveryDate: Date;
+		items: [PurchaseOrderItem];
+		attachments: [String];
+	}
+
+	# might belong in a different file but just gonna throw it here for now
+	type Person {
+		id: ID!;
+		name: String;
+		avatarHref: String;
```
MatthewGalan, line 55: For consistency throughout the schema, I think we should rename `avatarHref` to `image`.

MatthewGalan (changes requested):
Great start! Left a couple suggestions.

server/src/schemas/storeFrontSchema.ts
pjw9963: What would be in a label type? just a string?

server/src/schemas/storeFrontSchema.ts
pjw9963: Fair point

server/src/schemas/storeFrontSchema.ts
```diff
@@ -0,0 +1,74 @@
...
+		image: String;
+		name: String;
+		labels: [String];
+		unit: String;
+		pluralUnit: String;
+		count: Int;
+		pricePerUnit: Float;
+	}
+
+	type PurchaseOrderItem {
```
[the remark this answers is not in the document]
pjw9963: We could probably get away with just using `InventoryItems`, but I like to have the two distinct types since they represent different things.

server/src/schemas/storeFrontSchema.ts
MatthewGalan: I was thinking the `Label` could have an ID and a string name. That way, if an admin renames a label from the portal, we can just update the name property.

However, I think we could still manage this with the way you have it, and I think my way might be overcomplicating things, so let's just keep it as-is.

MatthewGalan (approved):

PR #25, opened 2021-11-16
fix the issue with graphql playground not working locally

pjw9963:
Helmet was messing with the content security policy. Adding the plugin makes the playground load when visiting localhost:port/graphql


MatthewGalan (approved):
LGTM but maybe get another approval too lol

PR #26, opened 2021-11-23
schema for machines

smellslikekeenspirit:


server/src/service/machinesSchema.ts
```diff
@@ -0,0 +1,75 @@
...
+    room: String!
+    addedAt: Date
+    inUse: Boolean!
+  }
+
+  type MachineFamily {
+    id: ID!
+    name: String!
+    description: String
+    trainingModule: Int
```
pjw9963, line 17: instead of having the `trainingModule` be an `Int` you have it be the actual `trainingModule` type

server/src/service/machinesSchema.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  type Reservation {
+    id: ID!
+    studentId: Int!
+    machineId: Int!
```
pjw9963, line 23: same thing as above. These Id fields really should just be actual type instead of an int. That way graphql can resolve everything the client asks for in a single request

pjw9963 (changes requested):
Looks pretty good. only thing is the "foreign-key" references on the types. I think you should just have the actual type there instead. I haven't used types from different files yet so let me know if its weird at all

server/src/service/machinesSchema.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  type MachineFamily {
+    id: ID!
+    name: String!
+    description: String
+    trainingModule: Int
+  }
+
+  type Reservation {
```
MatthewGalan, line 20: I thought we were making it to reservations had an associated expert as well?

server/src/service/machinesSchema.ts
smellslikekeenspirit: @pjw9963 in that case would the input have to be a full `trainingModule` object instead of just an identifier?

server/src/service/machinesSchema.ts
smellslikekeenspirit: like a labbie?

MatthewGalan (approved):

server/src/service/machinesSchema.ts
pjw9963: inputs cant have full types, only scalar values. so I would have the input have a trainingModule identifier to an existing training module

server/src/service/machinesSchema.ts
```diff
@@ -37,12 +38,13 @@ export const MachinesTypeDefs = gql`
...
     name: String!
     description: String
-    trainingModule: Int!
+    trainingModule: TrainingModule!
   }
 
   input ReservationInput {    
-    studentId: Int!
-    machineId: Int!
+    userId: User!
```
pjw9963, line 45: From the graphql docs: "Input types can't have fields that are other objects, only basic scalar types, list types, and other input types". So unless user and machine are inputs I don't think this works. I think it is fine to just use an int (primary key) here

pjw9963 (changes requested):
I like the changes you made to the types, just the inputs cant have types as fields

server/src/service/machinesSchema.ts
smellslikekeenspirit: @pjw9963 Yeah I was confused about this and then checked your inputs and you seem to have used types in them too. So we need to change these as well then? https://github.com/rit-construct-makerspace/makerspace/blob/71dd193f21bfa733f94340075d677dbf9b794e78/server/src/service/trainingSchema.ts#L38

pjw9963: that's an enum not a type so I think that's okay to use

PR #36, opened 2022-01-19
Add call to our GraphQL API

MatthewGalan:
This is a simple vertical-slice proof-of-concept in which our frontend makes a successful call to our backend.

Notes:
- Uses the `process.env.REACT_APP_GRAPHQL_URL` environment variable if provided, otherwise defaults to `http://localhost:3000/graphql`
- This means, in development, we can have our backend running on port 3000 and our frontend on port 3001. Currently, when starting the frontend while the backend is already running, you get a prompt saying that 3000 is already used and asking if you wanna use 3001. Say yes. I'll look into setting 3001 as the default, but it's not a high priority.

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -14,89 +14,97 @@ import StorefrontPage from "./pages/admin/storefront/StorefrontPage";
 import StorefrontPreviewPage from "./pages/admin/storefront_preview/StorefrontPreviewPage";
 import InventoryPage from "./pages/admin/inventory/InventoryPage";
 import EditMaterialPage from "./pages/admin/edit_material/EditMaterialPage";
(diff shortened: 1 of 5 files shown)


client/src/App.tsx
```diff
@@ -14,89 +14,97 @@ import StorefrontPage from "./pages/admin/storefront/StorefrontPage";
...
-              </Route>
-            </Switch>
-          </Box>
-        </Switch>
-      </BrowserRouter>
-    </ThemeProvider>
+    <ApolloProvider client={apolloClient}>
+      <ThemeProvider theme={theme}>
+        <BrowserRouter>
+          <Switch>
```
MatthewGalan, line 31: Didn't change any of the routes below. Might be worth bringing the routes into their own component so `App.tsx` is a bit cleaner.

PR #50, opened 2022-01-23
Extract route definitions to separate component

MatthewGalan:
Our frontend entry point, `App.tsx`, was getting a little cluttered with all the React Router route definitions. I moved these to their own component, which tidies things up nicely.


pjw9963 (approved):
👍

PR #56, opened 2022-01-25
prototype user management page

MatthewGalan:
Prototyped user management page. Clicking on a user card will (in theory) bring you to the management page for that specific user, where you can view more detailed information and perform management actions.

image

### DIFF
--- a/client/src/pages/admin/users/UserCard.tsx
+++ b/client/src/pages/admin/users/UserCard.tsx
@@ -1,0 +1,44 @@
+import React from "react";
+import {
+  Avatar,
+  Card,
+  CardActionArea,
+  Chip,
+  Stack,
+  Typography,
+} from "@mui/material";
+import User from "../../../types/User";
+
+interface UserCardProps {
+  user: User;
+}
+
+export default function UserCard({ user }: UserCardProps) {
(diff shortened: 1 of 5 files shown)


client/src/pages/admin/users/UserCard.tsx
```diff
@@ -0,0 +1,44 @@
...
+            variant="body1"
+            fontWeight={500}
+            width={120}
+            textAlign="center"
+            lineHeight={1.25}
+            flex={1}
+          >
+            {user.name}
+          </Typography>
+          {user.role !== "Maker" && (
```
pjw9963, line 32: I don't understand this. It looks like you are combining a boolean expression and a react component. Ignore me if this is fine

pjw9963 (approved):
Looks nice, left a comment on some code I didn't quite understand. Might be a bug, might be me not knowing syntax

client/src/pages/admin/users/UserCard.tsx
MatthewGalan: React docs explain it best: https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator

PR #59, opened 2022-01-26
Create holds migration

smellslikekeenspirit:
plus minor fixes and nits

### DIFF
--- a/server/src/db/migrations/20210923214317_initial_users.ts
+++ b/server/src/db/migrations/20210923214317_initial_users.ts
@@ -10,8 +10,8 @@ export async function up(knex: Knex): Promise<void> {
             t.string('first_name', 100);
             t.string('last_name', 100);
             t.text('email').unique();
-            t.enu('type', ['student', 'faculty', 'employee']);
-            t.enu('privilege_level', ['maker', 'labbie', 'admin']);
+            t.enu('type', ['STUDENT', 'FACULTY', 'EMPLOYEE']);
+            t.enu('privilege_level', ['MAKER', 'LABBIE', 'ADMIN']);
             t.date('registration_date').defaultTo(knex.fn.now());
             t.decimal('balance');
             t.integer('holds');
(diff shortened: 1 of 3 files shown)


server/src/db/migrations/20210923214317_initial_users.ts
```diff
@@ -10,11 +10,11 @@ export async function up(knex: Knex): Promise<void> {
...
             t.string('last_name', 100);
             t.text('email').unique();
-            t.string('type', 8); // student, faculty, employee
-            t.string('privilege_level', 6); // maker, labbie, admin
+            t.enu('type', ['STUDENT', 'FACULTY', 'EMPLOYEE']);
+            t.enu('privilege_level', ['MAKER', 'LABBIE', 'ADMIN']).defaultTo('MAKER');
             t.date('registration_date').defaultTo(knex.fn.now());
             t.decimal('balance');
-            t.boolean('hold').defaultTo(false);
+            t.integer('holds').defaultTo(0);
```
pjw9963, line 17: This holds field might not need to exist. Could be derived instead (join Holds table on the user Id), then we don't have to worry about this becoming out of sync

server/src/db/migrations/20220126063347_holds_table.ts
```diff
@@ -0,0 +1,27 @@
...
+
+
+export async function up(knex: Knex): Promise<void> {
+    return knex.schema.hasTable('Holds').then(function(exists) {
+        if (!exists) {
+            return knex.schema.createTable('Holds', function(t) {
+                t.increments('id');
+                t.enu('type', ['INCOMPLETE_TRAINING', 'SAFETY_VIOLATION', 'BALANCE_DUE']);
+                t.text('description');
+                t.integer('created_by').references('id').inTable('Users');
```
pjw9963, line 11: There might be situations where a hold is just automatically given to someone (I could be wrong but I vaguely remember a conversation), in that scenario the `created_by` field is a little awkward.

pjw9963 (changes requested):
I think getting rid of the holds field on the users table might be better. Also don't make it a habit to edit existing migrations, fine for now tho

PR #61, opened 2022-01-26
Create holds schema

smellslikekeenspirit:
Also, changed type of some `Date` fields in the machines schema to `DateTime` because the corresponding fields in the migration are `timestamp` fields. A GraphQL `DateTime` should be equivalent to a PostgreSQL `timestamp`.

### DIFF
--- a/server/src/schemas/holdsSchema.ts
+++ b/server/src/schemas/holdsSchema.ts
@@ -12,6 +12,7 @@ export const HoldsTypeDefs = gql`
     id: ID!
     type: HoldType!
     description: String
+    createdBy: User!
     createdAt: DateTime
     removedAt: DateTime
     userId: User!
@@ -21,6 +22,7 @@ export const HoldsTypeDefs = gql`
     id: Int!
     type: HoldType!
     description: String
+    createdBy: Int!
     createdAt: DateTime
     removedAt: DateTime
     userId: Int!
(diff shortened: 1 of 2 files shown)


MatthewGalan:
I like it. Should we add a `createdBy: User!` field as well? We wouldn't show the maker who placed the hold on their account, but it might be useful for admins to see how put the hold in the first place.

smellslikekeenspirit:
> I like it. Should we add a `createdBy: User!` field as well? We wouldn't show the maker who placed the hold on their account, but it might be useful for admins to see how put the hold in the first place.

yep good idea

server/src/schemas/holdsSchema.ts
```diff
@@ -21,6 +22,7 @@ export const HoldsTypeDefs = gql`
     id: Int!
     type: HoldType!
     description: String
+    createdBy: Int!
```
MatthewGalan, line 25: Why is this an int? Is it the user ID of the person who created it? If so, can we rename to something like `creatorId`?

smellslikekeenspirit: @MatthewGalan it does return a `User` object in the type. The corresponding input type requires an `Int` because we just need the ID of the person who created it.

MatthewGalan (approved):

PR #62, opened 2022-01-26
First try at a correct schema no laughing

Axj2689:


pjw9963 (changes requested):
bro this is an empty file

Axj2689:
I often forget that I never set up Autosave on VSCode. Fixed with file's actual contents and set up autosave so this doesn't happen again

server/src/schemas/usersSchema.ts
```diff
@@ -0,0 +1,73 @@
+import { gql } from "apollo-server-express";
+
+export const UsersTypeDefs = gql`
+  type User {
+    id: ID!
+    first_name: String!
+    last_name: String!
+    email: String!
+    type: Type!
```
MatthewGalan, line 9: Can we rename this to something more specific, like `userType`? "type" is a protected word in a lot of cases.

server/src/schemas/usersSchema.ts
```diff
@@ -0,0 +1,73 @@
...
+    privilege: Privilege!
+    registration_date: Date!
+    balance: Int!
```
MatthewGalan, line 12: I thought we weren't doing user balances?

server/src/schemas/usersSchema.ts
```diff
@@ -0,0 +1,73 @@
...
+    hold: Boolean!
+    holds: [Hold]
```
MatthewGalan, line 14: What's the point of `hold` when we have `holds`?

server/src/schemas/usersSchema.ts
```diff
@@ -0,0 +1,73 @@
...
+    training_modules: [TrainingModule]
+    year: Int
+    college: String
+    major: String
+    description: String
```
MatthewGalan, line 19: What sort of stuff would go in `description`?

server/src/schemas/usersSchema.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  enum Type{
+    STUDENT
+    FACULTY
+    EMPLOYEE
```
MatthewGalan, line 25: Is there a benefit to having a distinction between `FACULTY` and `EMPLOYEE`? I'm thinking an `isStudent` boolean might be simpler.

server/src/schemas/usersSchema.ts
```diff
@@ -0,0 +1,73 @@
...
+    email: String!
+    type: Type!
+    privilege: Privilege!
+    balance: Int!
+    hold: Boolean!
+  }
+
+  type Query {
+    user: User
+    users: [User]
```
MatthewGalan, line 58: I'm not a GraphQL expert but I'm pretty sure we don't need `users` here, just `user`.

@smellslikekeenspirit @pjw9963 - can you guys weigh in?

server/src/schemas/usersSchema.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  }
+
+  type Mutation {
+    addStudentUser(user: StudentUserInput): User
+    addFacultyUser(user: FacultyUserInput): User
+    updateStudentUser(user: StudentUserInput): User
+    updateFacultyUser(user: FacultyUserInput): User
+    #addTrainingModuleToUser(userID: ID, TrainingModule: module): User
```
MatthewGalan, line 67: What is the `#` for here?

MatthewGalan (changes requested):

server/src/schemas/usersSchema.ts
Axj2689: I guess you could just check if holds is empty yeah

server/src/schemas/usersSchema.ts
Axj2689: This was for the little blurb in the about me for labbies/admins.

server/src/schemas/usersSchema.ts
Axj2689: Commenting out that line in the schema until I confirmed it was alright to have it in there. I don't know if TrainingModule is a valid parameter. I'm thinking an ID for the training module might be more appropriate.

server/src/schemas/usersSchema.ts
Axj2689: Sounds right after doing more reading ngl

server/src/schemas/usersSchema.ts
Axj2689: Forgot tbh

server/src/schemas/usersSchema.ts
Axj2689: Removed as per is_student suggestion

server/src/schemas/usersSchema.ts
MatthewGalan: Gotcha. This is more a matter of preference but IMO `aboutMe` might be more self explanatory. Up to you.

server/src/schemas/usersSchema.ts
MatthewGalan: CC: @pjw9963

server/src/schemas/usersSchema.ts
pjw9963: If you can get list of users using just the `user` query then we don't need the `users` query, which I believe is how it works

MatthewGalan (approved):

PR #66, opened 2022-01-28
Audit logs Prototype Page

axg3073:
Prototype Page for Audit Logs


MatthewGalan (approved):
Awesome job, Andre! This is an excellent start to the audit logs page. One thing I would suggest is to set up your IDE to run Prettier on-save, just so formatting is consistent across the board.

I understand the search functionality still needs some work, but I think this PR is still worth merging now. We can continue to enhance this page with future PRs.

PR #69, opened 2022-01-28
Move equipment page components to their own folder

MatthewGalan:
- Moved `/pages/both/EquipmentPage.tsx` -> `/pages/both/equipment/EquipmentPage.tsx`
- Moved `/pages/both/MachineCard.tsx` -> `/pages/both/equipment/MachineCard.tsx`

This is now compliant with how we were structuring things for the other pages.


Axj2689 (approved):
simple enuf

PR #70, opened 2022-01-28
Set default React port to 3001

MatthewGalan:
If you're running both the backend and frontend at the same time, you'll get a warning that port 3000 is already in use by whichever half you started first.

Setting frontend to run on port 3001 to avoid the issue in the first place.


Axj2689:
simple enuf, good fix

Axj2689 (approved):

PR #80, opened 2022-02-01
Audit logs schema

axg3073:
Made Audit Logs Schema for Logging

Also made a quick fix for TextField input for inputting dates for Audit Logs Page (would be used for filtering)


Axj2689 (commented):
Queries?

Axj2689 (approved):

PR #82, opened 2022-02-03
Storefront backend

pjw9963:
Big PR due to small brain. I'll point out spots that I'm not sure about


server/package.json
```diff
@@ -23,6 +23,7 @@
     "dotenv": "^10.0.0",
     "express": "^4.17.1",
     "graphql": "^15.6.0",
+    "graphql-scalars": "^1.13.6",
```
pjw9963, line 26: this is for the graphql `date` scalar type

server/src/db/migrations/20220119232016_store-front.ts
```diff
@@ -0,0 +1,123 @@
...
+                t.text('unit');
+                t.text('pluralUnit');
+                t.integer('count');
+                t.float('pricePerUnit');
+            });
+        }
+    });
+
+    // if we know every label we can use an enum instead of this table
+    knex.schema.hasTable('Label').then(function (exists) {
```
pjw9963, line 21: Created a separate table instead of enum so that end-users can create their own specific labels. If we can think of every label they would ever need it might be better to just do an enum...

server/src/db/migrations/20220119232016_store-front.ts
```diff
@@ -0,0 +1,123 @@
...
+                    .onDelete('CASCADE');
+            });
+        }
+    });
+
+    knex.schema.hasTable('PurchaseOrder').then(function (exists) {
+        if (!exists) {
+            return knex.schema.createTable('PurchaseOrder', function (t) {
+                t.increments('id').primary();
+                t.text('creator'); // update this when we have user stuff done
```
pjw9963, line 48: Right now this is just a text field but should probably be some sort of userId

server/src/schemas/storeFrontSchema.ts
```diff
@@ -5,64 +5,73 @@ export const StoreFrontTypeDefs = gql`
...
-    count: Int;
+    id: ID!
+    item: InventoryItem
+    count: Int
   }
 
   type PurchaseOrder {
-    id: ID!;
-    creator: Person;
-    createDate: Date;
```
pjw9963, line 37: These are all dates at the moment, should they be datetimes instead?

server/src/service/InventoryService.ts
```diff
@@ -0,0 +1,35 @@
+import { InventoryItem } from "../models/store/inventoryItem";
+import { IInventoryRepo } from "../repositories/Store/inventoryRepository";
+
+export class InventoryService {
+    inventoryRepo: IInventoryRepo;
+   
+    constructor(repo: IInventoryRepo) {
+      this.inventoryRepo = repo;
+    }
+
```
pjw9963, line 10: These service classes seem a little pointless atm, but the idea is that if we need to multiple calls on a repo, any sort of logic, or error handling, I would rather have that code in here than cluttering the resolver

server/src/db/migrations/20220119232016_store-front.ts
```diff
@@ -0,0 +1,123 @@
+import { Knex } from "knex";
+
+
+export async function up(knex: Knex): Promise<void> {
+
+    knex.schema.hasTable('InventoryItem').then(function (exists) {
```
MatthewGalan, line 6: This is just a stylistic question, but is there a reason we're doing `function (exists) {` instead of `(exists) => {`?

server/src/mappers/store/InventoryItemMapper.ts
```diff
@@ -0,0 +1,28 @@
+import { InventoryItem } from "../../models/store/inventoryItem";
+
+export class InventoryItemMappper {
+  public static toDomain(raw: any): InventoryItem[] {
```
MatthewGalan, line 4: Is there a benefit to having it be a static function instead of just
``` ts
export default function toDomain(raw: any): InventoryItem[] {
```

server/src/resolvers/storeFrontResolver.ts
```diff
@@ -0,0 +1,91 @@
...
+      
+      // PurchaseOrder(Id: ID!): PurchaseOrder
+      PurchaseOrder: async (_: any, args: any, context: any) => {
+        return purchaseOrderService.getPurchaseOrder(args.Id);
+      }, 
+    },
+  
+    Mutation: {
+      // createPurchaseOrder(PO: PurchaseOrderInput): PurchaseOrder  
+      createPurchaseOrder: async (_: any, args: any) => {
```
MatthewGalan, line 34: Would this be possible?
``` ts
createPurchaseOrder: async (_: any, args: { PO: PurchaseOrder }): PurchaseOrder => {
```
I like that we have types in the comments above but I think it would be even better if we could write them as actual code.

server/src/service/PurchaseOrderService.ts
```diff
@@ -0,0 +1,53 @@
...
+import { IPurchaseOrderRepo } from "../repositories/Store/poRepository";
+
+export class PurchaseOrderService {
+  poRepo: IPurchaseOrderRepo;
+
+  constructor(repo: IPurchaseOrderRepo) {
+    this.poRepo = repo;
+  }
+
+  public async getAllPurchaseOrders(): Promise<PurchaseOrder[]> {
```
MatthewGalan, line 12: If all the service class does is call the repo, do we need it? Seems like a [middle man](https://refactoring.guru/smells/middle-man) to me, at least at first glance.

server/src/repositories/Store/poRepository.ts
```diff
@@ -0,0 +1,187 @@
...
+
+export class PurchaseOrderRepo implements IPurchaseOrderRepo {
+
+  private queryBuilder
+
+  constructor(queryBuilder?: any) {
+    this.queryBuilder = queryBuilder || knex
+  }
+
+  public async getAllPOs(): Promise<PurchaseOrder[]> {
```
MatthewGalan, line 30: Similarly, I don't really understand why these functions need to be part of a class, rather than just a series of exported functions. Is it just so we can do dependency injection on `queryBuilder`?

MatthewGalan (commented):
Great work! I just left a few questions regarding the broader design of our backend.

server/src/db/migrations/20220119232016_store-front.ts
MatthewGalan: Sounds good - maybe add a `TODO:` as part of the comment so we can find it easier later.

server/src/service/PurchaseOrderService.ts
MatthewGalan: Don't reply to this thread. Redirecting convo to https://github.com/rit-construct-makerspace/makerspace/pull/82#discussion_r798776165

server/src/repositories/Store/poRepository.ts
```diff
@@ -0,0 +1,187 @@
...
+  addAttachmentToPO(attachment: string, poId: number): Promise<PurchaseOrder>;
+  addAttachmentsToPO(attachments: string[], poId: number): Promise<PurchaseOrder>;
+  removeAttachmentFromPO(attachment: string, poId: number): Promise<PurchaseOrder>;
+  removeItemFromPO(itemId: number, poId: number): Promise<PurchaseOrder>;
+  updateItemCountInPO(itemId: number, count: number, poId: number): Promise<PurchaseOrder>;
+  deletePOById(poId: number): Promise<void>;
+}
+
+
+export class PurchaseOrderRepo implements IPurchaseOrderRepo {
```
MatthewGalan, line 22: Also can we rename this file to `purchaseOrderRepository` instead of `poRepository`

server/src/service/InventoryService.ts
MatthewGalan: That was my initial reaction, but I agree we're definitely going to need business logic and error handling in the near future. I'm becoming more on board with the whole service & repos pattern as I understand it more.

server/src/schemas/storeFrontSchema.ts
MatthewGalan: I think in general we should default to `DateTime` in the backend. If the user doesn't need to know the time of an event, that's something we can handle in the frontend.

server/src/repositories/Store/poRepository.ts
```diff
@@ -0,0 +1,187 @@
...
+      creator: creatorId,
+      createDate: new Date().toISOString(),
+      expectedDeliveryDate: expectedDeliveryDate,
+    }, "id"))[0];
+
+    if (items.length > 0) {
+     this.addItemsToPO(newId, items);
+    }
+
+    if (attachments.length > 0) {
```
MatthewGalan, line 101: Could `items.length > 0` and `attachments.length > 0` be considered business logic that should belong in the service layer instead?

server/src/db/migrations/20220119232016_store-front.ts
pjw9963: probably should add a way for people to create labels tho...

server/src/db/migrations/20220119232016_store-front.ts
pjw9963: I probably just copied this from Knex documentation. Arrow functions look a bit cleaner so I might update

server/src/mappers/store/InventoryItemMapper.ts
pjw9963: Nope, just not used to making functions without a class I guess.

server/src/resolvers/storeFrontResolver.ts
pjw9963: If this works I will 100% do this

server/src/repositories/Store/poRepository.ts
pjw9963: the `queryBuilder` parameter is for transactions.  I've been using a class for the repo so you could initialize a repo with a transaction object and then all performed methods would be apart of that one transaction. I was running into some limitations doing it this way though and was considering having an extra parameter on the methods for passing in an optional transaction object...

server/src/repositories/Store/poRepository.ts
pjw9963: eh maybe? Its just a check to see if the repo needs to insert attachments/items. I'll play around with it

MatthewGalan: That's fine, just trying to reach an understanding of what qualifies as business logic.

server/src/mappers/store/InventoryItemMapper.ts
MatthewGalan: In that case I would suggest going with exported functions, but it's not a deal breaker either way.

server/package.json
```diff
@@ -10,7 +10,7 @@
     "start:dev": "npm run build:dev",
     "build:dev": "nodemon 'src/server.ts' --exec 'ts-node' src/server.ts -e ts,graphql",
     "knex:migrate:make": "knex --migrations-directory src/db/migrations migrate:make -x ts",
-    "knex:migrate:latest": "knex --knexfile dist/db/knexFile.js migrate:latest",
+    "knex:migrate:latest": "npm run build && knex --knexfile dist/db/knexFile.js migrate:latest",
```
smellslikekeenspirit, line 13: much needed addition tbh

server/src/db/migrations/20220119232016_store-front.ts
```diff
@@ -0,0 +1,123 @@
...
+    knex.schema.hasTable('Label').then((exists) => {
+        if (!exists) {
+            return knex.schema.createTable('Label', (t) => {
+                t.increments('id').primary();
+                t.text('label');
+            });
+        }
+    });
+
+    knex.schema.hasTable('InventoryItemLabel').then((exists) => {
```
smellslikekeenspirit, line 30: Is a differentiation between Label and InventoryItemLabel necessary?

server/src/repositories/Store/purchaseOrderRepository.ts
```diff
@@ -0,0 +1,177 @@
...
+      this.addAttachmentsToPO(attachments, newId);
+    }
+    console.log(newId);
+    return this.getPOById(newId);
+  }
+
+  public async addItemToPO(item: InventoryItem, poId: number, count: number): Promise<PurchaseOrder> {
+    await this.queryBuilder("PurchaseOrderItem")
+      .insert({
+        item: item.id,
```
smellslikekeenspirit, line 101: also wondering, since the code is almost identical, if addItemToPo needs to be its own thing if we can just send a single item in an array to addItemsToPo

server/src/db/migrations/20220119232016_store-front.ts
pjw9963: its so that the user can create their own labels. Using a table for the labels instead of an enum

server/src/repositories/Store/purchaseOrderRepository.ts
pjw9963: Good ponint

MatthewGalan (approved):

PR #83, opened 2022-02-03
Users resolvers

Axj2689:
No implementation, still lots to do, but most of the work left is in Repo/Model. Will have to change the resolvers and the args passed in once the methods they reference are implemented.


server/src/resolvers/usersResolver.ts
```diff
@@ -0,0 +1,102 @@
...
+        updateStudentUser: async (_: any, args: any) => {
+          try {
+            const ur = new UserRepository();
+            await ur.updateUser(args)
+          } catch (e) {
+            console.log("Error", e)
+          }
+        },
+    
+        addTraining: async (_: any, args: any) => {
```
MatthewGalan, line 53: This would just be for admins' use, right? In normal circumstances, once you pass the training module, the backend should automatically mark it as complete for you, without an additional call from the frontend.

pjw9963: @MatthewGalan just for my understanding, you envision that a user will take the training quiz and at somepoint their answers are sent to the backend from the client. Then the backend checks the user's answers and if they passed it it'll "addTraining" to that user?

Axj2689: Yes I envisioned this for system/admin use only.

MatthewGalan (approved):

PR #84, opened 2022-02-03
Created Inventory Preview Page for a Maker.

axg3073:
Preview Page for Maker for inventory of materials.

Added 5th inventory item that has a 0 count and would display opaque under the materials section (for both Admin and Maker)


MatthewGalan (approved):
Code looks good to me! Also, let's rename the directory to `inventory_preview` (lower case) instead of `Inventory_Preview`, so it's consistent with the other pages' directories.

PR #87, opened 2022-02-08
Refactoring of machines schema and table + datetime resolver + fixes

smellslikekeenspirit:
fixes:
- there were mutations in both Users and Holds for adding and removing holds. Removed the duplicate from Holds because having this functionality in Users makes more sense
- changed case of variables in tables to camelCase for easier mapping of schema field to database field (for when we have better mappers)
- integrated datetime resolver 
- made input fields optional that have default values (e.g. createdAt)


server/src/db/migrations/20211102194503_machines_table.ts
```diff
@@ -3,37 +3,38 @@ import { Knex } from "knex";
...
-    return knex.schema.hasTable('MachineFamilies').then(function(exists) {
+    return knex.schema.hasTable('EquipmentLabels').then(function(exists) {
         if (!exists) {
-            return knex.schema.createTable('MachineFamilies', function(t) {
+            return knex.schema.createTable('EquipmentLabels', function(t) {
                 t.increments('id');
                 t.string('name', 50);
-                t.text('description');
-                t.integer('training_module').references('id').inTable('TrainingModule');
+                t.specificType('trainingModules', 'integer ARRAY')
```
pjw9963, line 11: Doesn't this fail 1st normal form? I think its a lot better practice to have a separate table to represent the multiple values

server/src/resolvers/datetimeResolver.ts
```diff
@@ -0,0 +1,7 @@
+import { GraphQLDateTime } from 'graphql-iso-date';
+
+const customScalarResolver = {
+  DateTime: GraphQLDateTime
+};
+
+export default customScalarResolver;
```
pjw9963, line 7: I think I imported a different datetime scalar from graphql-tools, we should probably use the same one

pjw9963 (changes requested):
Looks good, I don't like storing arrays in columns tho

server/src/resolvers/datetimeResolver.ts
pjw9963: Honestly I don't know the difference and they probably do the exact same thing. I think that all of our graphql dependencies are coming from graphql-tools at the moment tho so I'm leaning towards that one

PR #90, opened 2022-02-11
Audit log resolvers/Repo/Mapper

axg3073:
Worked on making the resolvers, repo, and mapper for Audit for the backend.

Unsure of my completion, may want a review unless all is good for now.


server/src/repositories/AuditLogs/AuditLogRepository.ts
```diff
@@ -0,0 +1,74 @@
...
+            .from("AuditLogs")
+            .where("id", logID);
+
+        return singleLogToDomain(knexResult);
+    }
+
+    public async getLogs(): Promise<AuditLogs[]> {
+        const knexResult = await this.queryBuilder("InventoryItem").select(
+            "AuditLogs.id",
+            "AuditLogs.timeDate",
```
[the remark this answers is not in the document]
axg3073: You mean like 36, that is 100% overlook on my part with "InventoryItem" instead of "AuditLogs"
fixing that now and going to overlook everything else one more time.

axg3073:
> Looks alright structure wise. You might want to double check your db code and make sure it works. Also we might want a way to select logs based on time, equipment or users as well

I'm actually uncertain how my db code should be and may want a quick review on that.

For selecting logs by time and users, I can add that. For equipment, that would been saying if equipment were a param.
I could do EventType which has "EQUIPMENT_MANAGEMENT" type and many more

MatthewGalan (approved):
LGTM but I'm not an authority on the backend stuff, so hoping @pjw9963 and/or @smellslikekeenspirit can submit an approval as well.

pjw9963:
> > Looks alright structure wise. You might want to double check your db code and make sure it works. Also we might want a way to select logs based on time, equipment or users as well
> 
> I'm actually uncertain how my db code should be and may want a quick review on that.
> 
> For selecting logs by time and users, I can add that. For equipment, that would been saying if equipment were a param. I could do EventType which has "EQUIPMENT_MANAGEMENT" type and many more

On the db code, It doesn't look like you've defined the auditlogs table yet. I think its a lot easier to write the implementation for your repository classes after you've decided how the database tables should be structured. Start by generating a migration using the script in the readme. That will generate a file with method stubs for adding/rolling back your changes. You can look at previous migrations and the knex website for help. Let me know if I'm not making sense and I can walk you through it (if you mic works lol)

server/src/repositories/AuditLogs/AuditLogRepository.ts
```diff
@@ -0,0 +1,120 @@
...
+            "AuditLogs.description"
+        );
+        return logsToDomain(knexResult);
+    }
+
+    public async addLog(log: AuditLogs): Promise<AuditLogs | null> {
+        const newID = (
+            await this.queryBuilder("AuditLogs").insert(
+                {
+                    timeDate: log.timeDate,
```
smellslikekeenspirit, line 96: If this is just an auto-generated field that specifies when the log was created, I don't think you should insert a value. You can make the db table add the current timestamp by default

smellslikekeenspirit:
I think you could also review the schema while you make the table/migration so that the two are consistent. The field called `timeDate` is using `Date` as type. You could use `DateTime` instead since the name suggests it's a timestamp (as it probably will be in the db table too). Also, the description should probably be optional in both your table and schema (it's not optional right now in the schema)

server/src/schemas/auditLogsSchema.ts
```diff
@@ -3,10 +3,10 @@ import { gql } from "apollo-server-express";
 export const AuditLogsTypeDefs = gql`
   type Log {
     id: ID!
-    timeDate: Date!
+    timeDate: DateTime!
```
smellslikekeenspirit, line 6: for this to work you need to import the DateTime scalar, see `machinesSchema` or `storeFrontSchema` maybe for reference

server/src/resolvers/auditLogsResolver.ts
```diff
@@ -0,0 +1,73 @@
+import {AuditLogRepo} from "../repositories/AuditLogs/AuditLogRepository";
+
+const alr = new AuditLogRepo();
+
+const AuditLogResolvers = {
+    Query: {
+        auditLogs: async (_: any, args: any, context: any) => {
+            try {
```
pjw9963, line 8: I would get rid of these try catch blocks, these stop the error messages from getting to the graphql playgound

server/src/schemas/auditLogsSchema.ts
```diff
@@ -21,14 +24,18 @@ export const AuditLogsTypeDefs = gql`
...
     eventType: EventType!
-    description: String!
+    description: String
   }
   
   type Query {
     auditLogs: [Log]
+    auditLog(logID: ID!): Log
+    auditLogsByUser(user: User!): [Log]
+    auditLogsByEventType(eventType: EventType!): [Log]
```
pjw9963, line 37: still might want to get logs on a specific piece of equipment

pjw9963 (approved):
What you have looks good, added a few notes that you might want to address

PR #92, opened 2022-02-11
Model entities for equipment backend

smellslikekeenspirit:
fragmenting what would have been an all-in-one equipment PR


pjw9963 (approved):
Looks pretty good except for maybe one thing. You might want to remove any array fields on your types. With them on your types you'll have to manually build the objects in your repo/mapper. If you remove them, you can set up resolver functions that will automatically build the object. The inventory item resolver has a good example with the labels field. The way you're doing it now will work but you might be making more work for yourself. Sorry if I suck at explaining

smellslikekeenspirit:
> Looks pretty good except for maybe one thing. You might want to remove any array fields on your types. With them on your types you'll have to manually build the objects in your repo/mapper. If you remove them, you can set up resolver functions that will automatically build the object. The inventory item resolver has a good example with the labels field. The way you're doing it now will work but you might be making more work for yourself. Sorry if I suck at explaining

Ah I guess that is what we are supposed to do differently this time haha

PR #98, opened 2022-02-13
Query for inventory items from the frontend

MatthewGalan:
As title suggests, the inventory page now queries for the inventory items, instead of displaying test data.

The item labels and images are still test data, though.


pjw9963 (approved):
I like the request wrapper component

PR #99, opened 2022-02-15
repo, mappers and resolvers for equipment backend

smellslikekeenspirit:
plus `TrainingModule` model and mapper


server/src/mappers/equipment/Equipment.ts
```diff
@@ -0,0 +1,19 @@
+import { Equipment } from "../../models/equipment/equipment";
+
+export function equipmentToDomain(raw: any): Equipment[] {
+  return raw.map((i: any) => {
```
MatthewGalan, line 4: Do we need to consider the possibility of `raw` being `null` or `undefined`?

server/src/mappers/equipment/EquipmentLabel.ts
```diff
@@ -0,0 +1,16 @@
+import { EquipmentLabel } from "../../models/equipment/equipmentLabel";
+
+export function equipmentLabelsToDomain(raw: any): EquipmentLabel[] {
+  return raw.map((i: any) => {
+    return singleEquipmentLabelToDomain(i);
+  })
```
MatthewGalan, line 6: Style suggestion:
``` js
return raw.map((i: any) => singleEquipementLabelToDomain(i));
```

MatthewGalan (approved):
Awesome work! I think we might need to add a reservation status at some point (pending, confirmed, cancelled), but this is a great start.

PS: Can we change the PR title so something more descriptive pls

server/src/mappers/equipment/Equipment.ts
smellslikekeenspirit: @MatthewGalan it calls a function `singleEquipmentToDomain` which takes care of that

MatthewGalan: Right, but wouldn't we get an error like `Cannot read properties of undefined (reading 'map')` on line 4 first?

PR #123, opened 2022-02-22
Added roomsSchema (to be modified as needed)

axg3073:
Small single schema PR

### DIFF
--- a/server/src/schemas/roomsSchema.ts
+++ b/server/src/schemas/roomsSchema.ts
@@ -1,0 +1,48 @@
+import { gql } from "apollo-server-express";
+
+export const RoomTypeDefs = gql`
+
+  type Room {
+    id: ID!
+    name: String!
+    equipmentList: [Equipment]
+    currentUsers: [User]
+    labbies: [User]
+    isOpen: Boolean!
+  }
+
+  input RoomInput {
+    name: String!
+    equipmentList: [Equipment]
+    currentUsers: [User]
+    labbies: [User]
+    isOpen: Boolean!
+  }
+  
+  type Query {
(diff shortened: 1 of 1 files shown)


server/src/schemas/roomsSchema.ts
```diff
@@ -0,0 +1,48 @@
...
+    labbies: [User]
+    isOpen: Boolean!
+  }
+  
+  type Query {
+    rooms: [Room]
+    room(roomID: ID!): Room
+    roomsByEquipment(equipment: Equipment!): [Room]
+    roomsByLabbie(labbie: User!): [Room]
+    roomsByIfOpen(isOpen: Boolean!): [Room]
```
MatthewGalan, line 27: Do you have a use case in mind for `roomsByEquipment`, `roomsByEquipment`, and `roomsByIfOpen`?

MatthewGalan (approved):
Awesome job! Looking very clean. Just had one question, otherwise LGTM.

server/src/schemas/roomsSchema.ts
axg3073: If normal makers are able to view different rooms through the portal, finding a room by Equipment and if the rooms is open useful to them
`ByEquipment`: Able to locate which rooms a specific equipment is located.
`ByIfOpen`: If it is possible to have a room closed off to users (fully reserved or closed due to circumstances/maintenace/afterhours/etc.) 
Only useful if keep data of Boolean value `ifOpen` for rooms

`ByLabbie`: Able to pull specific list of rooms of which a labbie user is monitoring. Only seems useful if getting the current data of what Labbie is currently monitoring which rooms.

PR #124, opened 2022-02-22
Added inventory queries to other pages

MatthewGalan:
Now the storefront page and maker inventory preview page also query for the real data. Also I added functionality to the search bars on the storefront, admin inventory, and maker inventory pages.


pjw9963 (approved):
N i c e

PR #125, opened 2022-02-23
Initial reservations schema

MatthewGalan:
Created the initial reservations schema. Let me know if I missed anything (or messed something up lol).

I found it helpful to look at the [Figma mockup](https://www.figma.com/file/XvgQbkldo8ojX4M2TwucLC/Scuffed?node-id=0%3A1) while writing the schema, so maybe it's helpful during review, too.

Not going to go into an in-depth explanation here because my hope is that the schema is self explanatory, otherwise I will tweak it to be more clear. :)

### DIFF
--- a/server/src/schemas/reservationsSchema.ts
+++ b/server/src/schemas/reservationsSchema.ts
@@ -7,7 +7,7 @@ Potential future enhancements:
   - Reschedule reservations
  */
 
-export const ReservationsSchema = gql`
+export const ReservationsTypeDefs = gql`
   enum ReservationStatus {
     PENDING
     CONFIRMED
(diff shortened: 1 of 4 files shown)


smellslikekeenspirit (commented):
Can you also remove the existing schema I had made under `machinesSchema`?

server/src/schemas/reservationsSchema.ts
```diff
@@ -0,0 +1,66 @@
...
+    CONFIRMED
+    CANCELLED
+  }
+
+  type Reservation {
+    id: ID!
+    maker: User
+    labbie: User
+    createDate: DateTime
+    creator: Person
```
smellslikekeenspirit, line 22: Tangential but why do we have both `User` and `Person` objects if a `User` is generic enough to be used for anyone in the system? Also we'll have their `type` and `privilegeLevel` known as well whereas the `Person` object as it is implemented stores very little information about the person. Is it for unregistered students to be able to make reservations? I thought Mike said they need to register first even if it's a walk-in but I might be wrong

smellslikekeenspirit:
Solid stuff

server/src/schemas/reservationsSchema.ts
MatthewGalan: My bad - I meant to do `creator: User` here. I agree that `Person` is redundant now. I pinged @pjw9963 and he agrees, so we can remove the `Person` type in a future PR.

MatthewGalan:
@smellslikekeenspirit Thank you for the review! I made the following changes:
- Replaced input `ID`s with `Int`s
- Removed old reservation schema
- Replaced `Person` with `User`

Pls take another look.

smellslikekeenspirit (approved):
LGTM 🚀

PR #126, opened 2022-02-23
Thanos snap purchase orders

MatthewGalan:
Removed purchase order code from our entire stack. Only thing remaining is a couple references in the migrations and tests - hoping @pjw9963 can lend his expertise there.

Bear in mind that this code will still be recoverable if we need it in the future, with the magic of version control. I just don't think it's wise to keep all this dead code around.

🙏 Rest in piece.


pjw9963 (approved):
Rip purchase orders, I can do a migration to get rid of the db schema in a bit

smellslikekeenspirit:
Innalillahi wa inna ilaihi rajiun (Muslim for RIP)

smellslikekeenspirit (approved):

PR #129, opened 2022-02-23
fixed inventory tests and removed purchase orders from the db schema

pjw9963:
Tests had references to Purchase Orders and were missing the threshold value


MatthewGalan (approved):
LGTM

PR #131, opened 2022-02-24
Frontend inventory enhancements

MatthewGalan:
_NOW THAT'S WHAT I CALL VELOCITY_

- You can now set the threshold for an item from the UI
- Items whose count falls below the threshold now show up in the "Running Low" section
- Item creation & updating is now a modal instead of a page
- Now one step closer to using Apollo caching properly (its hard, ok?)


client/src/pages/admin/inventory/InventoryPage.tsx
```diff
@@ -5,38 +5,46 @@ import SearchBar from "../../../common/SearchBar";
...
 
   const [searchText, setSearchText] = useState<string>("");
+  const [modalItemId, setModalItemId] = useState<string>("");
+
+  const { loading, error, data } = useQuery(GET_INVENTORY_ITEMS);
 
-  const { loading, error, data } = useQuery(QUERY_INVENTORY_ITEMS, {
-    fetchPolicy: "no-cache",
-  });
+  console.log(data);
```
pjw9963, line 27: Probably could get rid of this

pjw9963 (approved):
LGTM

PR #136, opened 2022-02-24
Innalillah labels

smellslikekeenspirit:


server/src/db/migrations/20211102194503_machines_table.ts
```diff
@@ -3,25 +3,7 @@ import { Knex } from "knex";
...
-              });
-        }
-    }).then(async () => {
-      // table that stores training modules attached to equipment labels
-      const exists = await knex.schema.hasTable('ModulesForLabels');
-      if (!exists) {
-        return knex.schema.createTable('ModulesForLabels', function (t) {
-          t.increments('id');
-          t.integer('equipmentLabelId').references('id').inTable('EquipmentLabels');
-          t.integer('trainingModuleId').references('id').inTable('TrainingModule');
```
MatthewGalan, line 20: Shouldn't we generally be creating new migrations instead of modifying existing ones?

CC: @pjw9963

MatthewGalan (commented):
Innalillahi wa inna ilaihi rajiun 🙏

server/src/db/migrations/20211102194503_machines_table.ts
pjw9963: Yeah. Modifying old migrations will break anyone's test db setup. That isn't too bad right now cuz generating a new db isn't too hard. But once we have any sort of production running db we really dont want to lose all the data in it

smellslikekeenspirit: Let me do this for now

MatthewGalan (approved):

PR #138, opened 2022-02-28
added mock auth and fixed issues preventing build on master

pjw9963:
Enabled by setting the `MOCK_AUTH` env value to `TRUE`, this will expose an endpoint at `/auth/mock` that will always auth the user. The user to be authenticated and inserted into the request is the `customUserObject` inside the `setupMockAuth` function. Change this object to suit your needs.

also fixed a few issues in master that were causing failures.

also run `npm i` cuz I added a mock auth package


server/src/auth.ts
```diff
@@ -17,7 +18,48 @@ declare global {
...
+      console.log('Mock Authenticating')
+      res.redirect('/')
+    });
+
+  passport.serializeUser(function (user, done) {
+    done(null, user);
+  });
+
+  passport.deserializeUser(function (user, done) {
+    // @ts-ignore
```
pjw9963, line 57: lots of these cuz the library will only accept a specific user type that does not match our needs. Let me know if there is a better way

MatthewGalan (approved):

PR #140, opened 2022-03-01
Fullstack training modules

MatthewGalan:
- Admin's training module page now queries and lists the training modules
- Clicking a training module brings you to the editor page for that module
- You can rename and delete the module
- You can add items to the module, including questions (with options), text, YouTube embeds, and images
- All modifications to the module make mutations as you go. The status of these mutations is shown in a Google Docs style "Saving..." "Saved" indicator.
- Refactorings & bug fixes in the training module related mappers and resolvers


server/src/repositories/Training/questionRepo.ts
```diff
@@ -1,67 +1,54 @@
...
 
-export interface IQuestionRepo {
-  getQuestionById(questionId: number | string): Promise<Question>;
-  getQuestionsByModule(moduleId: number | string): Promise<Question[]>;
-  addQuestionToModule(
-    moduleId: number | string,
-    question: Question
-  ): Promise<Question>;
-  deleteQuestionById(questionId: number | string): Promise<void>
+export async function getQuestion(
```
MatthewGalan, line 5: @pjw9963 I refactored the question and module repos to be a series of exported functions, instead of classes. IMO we weren't getting any benefit from having them as classes, and it was just adding boilerplate. LMK what you think of this instead.

server/src/db/migrations/20211010154323_build-training-tables.ts
```diff
@@ -1,65 +1,60 @@
...
+    if (!exists) {
+      return knex.schema.createTable("Question", function (t) {
+        t.increments("id").primary();
+        t.integer("module").references("TrainingModule.id").onDelete("CASCADE"); // If module is deleted, delete question as well
+        t.enu("questionType", [
+          "MULTIPLE_CHOICE",
+          "CHECKBOXES",
+          "TEXT",
+          "YOUTUBE",
+          "IMAGE",
```
MatthewGalan, line 23: my turn to modify existing migrations :P

server/src/repositories/Training/questionRepo.ts
pjw9963: Yeah this is fine. I had them as classes for testing and handling transactions but we are not doing either of those so 👍

server/src/models/equipment/equipment.ts
```diff
@@ -1,12 +1,7 @@
...
-
 export interface Equipment {
-    id: number;
-    name: string;
-    room: string;
-    addedAt: Date,
-    inUse: boolean;
+  id: number;
+  name: string;
+  room: string;
```
axg3073, line 4: Just make sure to fix this since room is now roomID and an int instead of string
```roomID: int
  room: string;
```

axg3073 (approved):
LGTM, there was a lot so mainly looked through the backend stuff. So far just that one edit that should be causing the merge conflict

server/src/models/equipment/equipment.ts
MatthewGalan: Fixed with merging master into my branch 👍

pjw9963 (approved):
Backend changes LGTM

PR #150, opened 2022-03-05
Create backend pipeline

MatthewGalan:
This is a starter CI pipeline for our backend. Right now it just starts the DB container and runs migrations, but it will serve as a good start.


MatthewGalan:
Merging so I can try the pipeline lol

PR #152, opened 2022-03-05
Storefront backend cleanup

MatthewGalan:
- Changed inventory and label repos to be exported functions instead of classes
- Moved TS interfaces for the storefront from `models/Store` to be in `storeFrontSchema` instead
  - If we continue this pattern, we could do away with the `models` directory entirely
  - IMO the TS interfaces are essentially an extension of the schema. If they're in the same file as the schema, then it will be easy to remember to update them when we update the schema


pjw9963 (approved):
LGTM, as long as our "models" just represent whats in our schema I think this is a good change

PR #153, opened 2022-03-05
Update inventory counts upon checkout

MatthewGalan:
When you checkout, the app now calls the `removeItemAmount` mutation for each item in the cart.

In the future it will probably be better to have a mutation that can handle multiple items at once so we don't have to make so many network requests.

However, I expect most carts will only have one or two items in them anyway, so this should be good enough for MVP.

---

**EDIT:** I had to nuke `package-lock.json` because we changed `npm i` to `npm ci` in the frontend pipeline and shit broke but this fixed it idk whats going on anymore


pjw9963 (approved):
nice

PR #154, opened 2022-03-07
Connect equipment management UI to backend

MatthewGalan:
It's now possible to create and modify equipment from the frontend! Also included some bugfixes in the backend.


server/src/resolvers/roomsResolver.ts
```diff
@@ -1,125 +1,123 @@
...
+  Query: {
+    rooms: async (_: any, args: any, context: any) => {
+      try {
+        return await rr.getRooms();
+      } catch (e) {
+        console.log("Error:", e);
+      }
+    },
+    room: async (_: any, args: any, context: any) => {
+      try {
```
pjw9963, line 19: These try catches might stop the Apollo client from knowing that any error happened which could be bad (like receiving null for rooms)

pjw9963 (approved):
React code looks nice. Good fix for the migration issue. Left a comment about using try/catches in resolvers, I figure you know about that tho and you were just fixing semantics. I'll have a PR shortly with some back-end training refactoring so don't worry about that atm

server/src/resolvers/roomsResolver.ts
MatthewGalan: Yup, agreed. I only touched this file to fix what I needed and was trying to keep this PR from growing out of control so I didn't refactor other stuff. Rest of the changes were Prettier reformatting.

PR #162, opened 2022-03-10
Manage users frontend

MatthewGalan:
- Manage users page now queries and displays the users
- Clicking on a user opens a modal instead of a page (but you'll never get me to admit to chow he was right)
- This modal queries and displays information about that user
- You can update the user's privilege from this modal


pjw9963 (approved):
LGTM !

PR #167, opened 2022-03-13
Training item order

pjw9963:
Refactored training back end to match implementation style of other sub systems. Also added order to training module questions.


server/src/schemas/trainingSchema.ts
```diff
@@ -1,5 +1,30 @@
...
+}
+
+export enum QuestionType {
+  MULTIPLE_CHOICE,
+  CHECKBOXES
+}
+
+export interface Question {
+  id: number | undefined;
+  moduleId?: number;
```
MatthewGalan, line 15: Do we need `Question.moduleId`?

server/src/schemas/trainingSchema.ts
```diff
@@ -1,5 +1,30 @@
...
+  text: string;
+  type: QuestionType;
+  order: number;
+}
```
MatthewGalan, line 19: Looks to me like `Question` is missing `options: Option[];` or similar

server/src/schemas/trainingSchema.ts
```diff
@@ -1,5 +1,30 @@
 import { gql } from "apollo-server-express";
 
+export interface TrainingModule {
+  id: number;
+  name: string;
+}
+
+export enum QuestionType {
+  MULTIPLE_CHOICE,
+  CHECKBOXES
```
MatthewGalan, line 10: I suggest we rename `QuestionType` to `ModuleItemType` or similar [(see frontend version)](https://github.com/rit-construct-makerspace/makerspace/blob/master/client/src/types/Module.ts#L1-L7). This is because they aren't necessarily questions. They can be information text blocks, YouTube embeds, or images.

MatthewGalan (approved):
LGTM with some minor comments. Lots of awesome refactoring here. Great work!

server/src/db/migrations/20220307225738_add-order-to-training-mod-items.ts
```diff
@@ -0,0 +1,15 @@
+import { Knex } from "knex";
+
+
+export async function up(knex: Knex): Promise<void> {
+    return knex.schema.alterTable("Question", function (t) {
```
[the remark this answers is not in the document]
pjw9963: yeah good catch

server/src/schemas/trainingSchema.ts
pjw9963: I needed the moduleID to update the order of all the related questions, I think there's a better way to do this tho

server/src/schemas/trainingSchema.ts
pjw9963: Our repos dont actually return related collections, graphql resolves it for us instead. It makes our repos and mappers much simpler to implement/maintain

server/src/schemas/trainingSchema.ts
pjw9963: Yeah I saw a comment about this, will update

pjw9963:
Added the `Question` to `ModuleItem` refactoring to this pr. @MatthewGalan  might want to update the frontend since the training schema changed slightly

MatthewGalan:
@pjw9963 Does this need another review then?

pjw9963:
Yeah, probably a good idea. Just make sure I didn't do something silly

server/src/repositories/Training/ModuleItemRepository.ts
```diff
@@ -0,0 +1,99 @@
...
+export async function getModuleItemsByModule(
+  moduleId: number | string
+): Promise<ModuleItem[]> {
+  const knexResult = await knex("ModuleItem")
+    .select(
+      "id",
+      "module",
+      "moduleItemType",
+      "text",
+      "order"
```
MatthewGalan, line 29: This may be a controversial take but I've recently taken to the idea of not including the column names and just doing `select()`. Since we're always feeding the data through a mapper, I don't think it's needed to always be explicit about which columns.

server/src/repositories/Training/ModuleItemRepository.ts
```diff
@@ -0,0 +1,99 @@
...
+    { module: moduleId, moduleItemType: moduleItem.type, text: moduleItem.text, order: newOrder },
+    "id"
+  );
+
+  return getModuleItem(insert[0]);
+}
+
+export async function updateModuleItemOrder(moduleId: number, moduleItemId: number, newOrder: number) {
+  let moduleItem = await getModuleItem(moduleItemId)
+  if (moduleItem === undefined || moduleItem === null)
```
MatthewGalan, line 52:
```suggestion
  if (!moduleItem)
```

server/src/repositories/Training/ModuleItemRepository.ts
```diff
@@ -0,0 +1,99 @@
...
+    (await knex("ModuleItem")
+      .select(
+        "module",
+        "order"
+      )
+      .where("id", id))[0];
+  await knex("ModuleItem")
+    .where({ id: id })
+    .update({ moduleItemType: moduleItem.type, text: moduleItem.text });
+  if (current.order !== moduleItem.order && current?.module != null)
```
MatthewGalan, line 83:
```suggestion
  if (current.order !== moduleItem.order && current?.module)
```

MatthewGalan (approved):
v nice

server/src/repositories/Training/ModuleItemRepository.ts
pjw9963: That makes sense 🧠

PR #168, opened 2022-03-16
Integrate w/ samltest.id & create DB entry for first timers

MatthewGalan:
- Removed `major` column from `Users` table and added `setupComplete` and `ritUsername` columns
- Mild refactoring on server initialization
- All but the final step of this sequence diagram has been implemented now!

### DIFF
--- a/server/src/schemas/usersSchema.ts
+++ b/server/src/schemas/usersSchema.ts
@@ -11,6 +11,7 @@ export enum Privilege {
 export interface User {
   id: number;
   universityID: string;
+  ritUsername: string;
   firstName: string;
   lastName: string;
   email: string;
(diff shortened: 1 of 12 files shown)


server/src/schemas/usersSchema.ts
```diff
@@ -58,6 +59,13 @@ export const UsersTypeDefs = gql`
     room: Room
     roomMonitoring: Room
 
+    """
+    The number-letter combination that is attached to your RIT email
+    (ie. abc1234). Not sensitive info. Stored plainly.
+    Not to be confused with the universityID.
```
Axj2689, line 65: nice

Axj2689 (approved):

PR #170, opened 2022-03-17
Signup form fiesta

MatthewGalan:
- Added signup form page
- You are redirected to this page if `setupComplete` is false for you
- You can fill in and submit this form, it's connected to the backend
- Left nav changes appearance based on your role (admin section is hidden for makers)
- Pronouns now displayed on user modal

### DIFF
--- a/client/src/common/CurrentUserProvider.tsx
+++ b/client/src/common/CurrentUserProvider.tsx
@@ -12,7 +12,9 @@ const GET_CURRENT_USER = gql`
       id
       firstName
       lastName
+      email
       privilege
+      setupComplete
     }
   }
 `;
@@ -26,9 +28,19 @@ interface CurrentUserProviderProps {
 export function CurrentUserProvider({ children }: CurrentUserProviderProps) {
   const result = useQuery(GET_CURRENT_USER);
 
-  // If the current user is null, we need to redirect to SSO login
+  // If the current user is null, redirect to SSO login
   if (!result.loading && !result.data.currentUser) {
     window.location.replace(loginUrl);
+  }
+
+  // If the user exists but setupComplete is false,
+  // redirect to them to the signup form
+  if (
+    result.data?.currentUser &&
+    !result.data.currentUser.setupComplete &&
+    !window.location.href.includes("/signup")
+  ) {
+    window.location.replace("/signup");
   }
 
   return (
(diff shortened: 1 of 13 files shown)


client/src/common/CurrentUserProvider.tsx
```diff
@@ -26,11 +28,21 @@ interface CurrentUserProviderProps {
...
   }
 
+  // If the user exists but setupComplete is false,
+  // redirect to them to the signup form
+  if (
+    result.data?.currentUser &&
+    !result.data.currentUser.setupComplete &&
+    !window.location.href.includes("/signup")
+  ) {
+    window.location.replace("/signup");
```
pjw9963, line 43: this might break the web browsers back button

server/src/repositories/Users/UserRepository.ts
```diff
@@ -41,8 +41,19 @@ export async function createUser(user: {
   return await getUserByID(newID);
 }
 
-export function updateUser(args: any) {
-  throw new Error("Method not implemented.");
+export async function updateStudentProfile(args: {
```
pjw9963, line 44: why single parameter? Seems fine just curious

server/src/resolvers/usersResolver.ts
```diff
@@ -19,12 +18,16 @@ const UsersResolvers = {
...
+    updateStudentProfile: async (
+      parent: any,
+      args: {
+        userID: number;
+        pronouns: string;
+        college: string;
+        expectedGraduation: string;
+      }
+    ) => {
+      return await UserRepo.updateStudentProfile(args);
```
pjw9963, line 30: I see why you use single parameter now lol

client/src/common/CurrentUserProvider.tsx
MatthewGalan: I think it's usually preferred to use `replace` for redirects. Since it wasn't the user doing the navigation, it shouldn't go onto the history stack. I can change this to `assign` tho if you would prefer.

server/src/repositories/Users/UserRepository.ts
MatthewGalan: Any time I have several arguments of the same type I prefer to go with an argument object so I don't have to memorize the argument order lol

PR #180, opened 2022-03-22
Fix type error on master

pjw9963:
our user object did not match the expressjs user object on the request and caused a type error.


MatthewGalan (approved):
mb I might've been responsible for this

PR #182, opened 2022-03-23
Lots of frontend chores

MatthewGalan:
- Migrated away from `react-app-rewired` to just vanilla `create-react-app`.
  - `react-app-rewired` is a deprecated configuration layer on top of `create-react-app`. We were only using it because the MUI documentation recommended it.
  - Sticking with plain old `create-react-app` will make troubleshooting in the future much easier.
- Added a favicon and set the page title to `Construct Portal`
- Started adding frontend tests. More to come. They run on the pipeline too.
- Deleted some old test data


pjw9963 (approved):
testing is cool 😎

pjw9963 (approved):
Routing is cool too 😎

PR #188, opened 2022-03-25
Better auth function

MatthewGalan:
I've replaced `context.userHasPrivilege()` with `context.ifAllowed()`.

`ifAllowed()` takes a callback that is executed _if and only if_
1. The user is authenicated
2. The user's privilege is included in the array of required privileges
3. The user has no holds (in a future enhancement)

Benefits:
1. If any of the above conditions are not met, an appropriate error will be thrown. The resolvers no longer have to handle this.
2. When the callback is executed, the user object is passed as an argument, and it is guaranteed to exist.

Looking forward to hearing your thoughts on this pattern. I hope it's easy to read/understand.

### DIFF
--- a/server/src/context.ts
+++ b/server/src/context.ts
@@ -1,0 +1,38 @@
+import { Privilege, User } from "./schemas/usersSchema";
+import { AuthenticationError, ForbiddenError } from "apollo-server-express";
+
+export interface ApolloContext {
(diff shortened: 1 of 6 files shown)


pjw9963 (approved):
I like this lots. Thanks for examples and moving the context logic into its own space

PR #189, opened 2022-03-25
Holds schema + interfaces

smellslikekeenspirit:


server/src/schemas/holdsSchema.ts
```diff
@@ -1,38 +1,26 @@
...
-  enum HoldType {
-    INCOMPLETE_TRAINING
-    SAFETY_VIOLATION
-    BALANCE_DUE
-  }
-
   type Hold {
     id: ID!
-    type: HoldType!
+    userID: User!
```
MatthewGalan, line 6:
```suggestion
    user: User!
```

server/src/schemas/holdsSchema.ts
```diff
@@ -1,38 +1,26 @@
...
-    createdAt: DateTime
-    removedAt: DateTime
-    userId: User!
+    active: Boolean
   }
 
   input HoldInput {
-    id: Int!
-    type: HoldType!
+    userID: Int!
```
MatthewGalan, line 12:
```suggestion
    userID: ID!
```

server/src/schemas/holdsSchema.ts
```diff
@@ -1,38 +1,26 @@
...
     description: String
-    createdBy: Int!
-    createdAt: DateTime
-    removedAt: DateTime
-    userId: Int!
+    active: Boolean
```
MatthewGalan, line 14: I don't think we need `active` in the input, since we'll always want it to be `true` when creating a new hold.

server/src/schemas/holdsSchema.ts
```diff
@@ -1,38 +1,26 @@
...
   }
 
   type Query {
     holds: [Hold]
+    holdsByUser(userID: ID!): [Hold]
```
MatthewGalan, line 19: IMO we can get rid of both of these queries, since we can just query `user.holds` instead.

server/src/models/holds/holdInput.ts
```diff
@@ -0,0 +1,6 @@
+
+export interface HoldInput {
+    userID: number;
+    description: string;
+    active: boolean;
+}
```
MatthewGalan, line 6: Can we put both this and the `Hold` interface in `holdsSchema.ts` instead, like we're doing with the users schema?

MatthewGalan (changes requested):

server/src/schemas/holdsSchema.ts
smellslikekeenspirit: I don't think `user.holds` which returns an array should exist. This is actually like a lookup table for holds which is better imo. @MatthewGalan I think we should segregate holds functionality from the Users stuff too

smellslikekeenspirit: I added the `userID` field so that we no longer need to keep track of holds in `Users`

MatthewGalan: IMO the API design should be independent from the database design. Conceptually speaking, holds are attached to users, and thus should be accessed via `user.holds`. `holdsByUser` feels like something from a REST API, not a GraphQL API.

CC: @pjw9963 thoughts?

pjw9963: I think a hold is always connected to a user, making it like a property. Graphql connects our properties like a graph, so the end-user can query for a user object and select whatever properties they want as long as they are related. I think separating the logic and db implementation of users and holds is good but the graphql resolver should tie them together and allow for the end-user to query for a user and their corresponding holds.

server/src/schemas/holdsSchema.ts
MatthewGalan: Do you disagree with this suggestion? The conversation was marked as resolved but the change wasn't implemented.

server/src/schemas/holdsSchema.ts
MatthewGalan: Same here?

server/src/schemas/holdsSchema.ts
MatthewGalan: ...and here?

smellslikekeenspirit: @MatthewGalan my bad I forgot to commit

MatthewGalan (approved):

PR #190, opened 2022-03-27
Signup & module browser tests

MatthewGalan:
I heckin love tests!


Axj2689 (approved):
Really like the testing through user-event instead of testing the method calls directly.

PR #191, opened 2022-03-27
Update reservations (Scheduling work in progress)

smellslikekeenspirit:


server/src/repositories/Equipment/ReservationRepository.ts
```diff
@@ -6,10 +6,14 @@ import { reservationsToDomain, singleReservationToDomain } from "../../mappers/e
 export interface IReservationRepository {
   getReservationById(id: number | string): Promise<Reservation | null>;
   getReservations(): Promise<Reservation[]>;
-  addReservation(reservation: ReservationInput): Promise<Reservation | null>;
+  createReservation(reservation: ReservationInput): Promise<Reservation | null>;
+  assignLabbieToReservation(resId: number, labbieId: number): Promise<Reservation | null>;
+  removeLabbieFromReservation(resId: number): Promise<Reservation | null>;
+  addComment(resId: number, authorId: number, commentText: string): Promise<string | null>;
```
MatthewGalan, line 12: I don't think we need `authorID` since we can get that from the context.

server/src/models/equipment/reservationInput.ts
```diff
@@ -1,8 +1,10 @@
 export interface ReservationInput {
-    userId: number;
-    supervisorId: number;
-    machineId: number;
-    createdAt: Date;
+
+    creator: number;
```
MatthewGalan, line 3: I don't think we need `creator` since we can get that from the context.

server/src/repositories/Equipment/ReservationRepository.ts
```diff
@@ -54,32 +64,129 @@ export class ReservationRepository implements IReservationRepository {
...
+      });
+      // in progress
+      // const startTime = reservation.startTime;
+      // const endTime
+      // const free = await this.queryBuilder("Reservations")
+      //   .select("*")
+      //   .whereBetween("startTime", [reservation.startTime, reservation.endTime])
+      //   .whereBetween("endTime", [reservation.startTime, reservation.endTime])
+      // if maker has passed all modules 
+      if (passed) {
```
MatthewGalan, line 107: I think in general we want to limit the repositories to simple database operations, and put the business logic in the resolvers.

CC: @pjw9963 for confirmation

server/src/repositories/Equipment/ReservationRepository.ts
```diff
@@ -54,32 +64,129 @@ export class ReservationRepository implements IReservationRepository {
...
+          user: authorId,
+          payload: commentText
+        },
+        "id"
+      )
+    )[0];
+    
+    // return last added comment
+    return knex("Reservations")
+    .join(
```
MatthewGalan, line 159: Do we need this? Can't we just return the comment ID instead?

server/src/resolvers/reservationResolver.ts
```diff
@@ -0,0 +1,51 @@
...
+  Query: {
+    reservations: async (_: any, args: { Id: number }, context: any) => {
+      return await reservationRepo.getReservations();
+    },
+
+    reservation: async (_: any, args: { Id: number }, context: any) => {
+      return await reservationRepo.getReservationById(args.Id);
+    },
+
+    reservationByUser: async (_: any, args: { Id: number }, context: any) => {
```
MatthewGalan, line 17: We had a similar conversation about holds - IMO doing `user.reservations` is more GraphQL friendly than `reservationsByUser`.

CC: @pjw9963

MatthewGalan (changes requested):
Good work! A few points:
- pls replace all instances of `somethingId` with `somethingID`, and `Id` with `id` for consistency
- whenever we have an object property that is just an ID, imo we should call it `makerID` rather than `maker`, for example
- I don't think we need to add the ability to delete/archive reservations, as just cancelling them should be sufficient. I assume Mike will still want to be able to view cancelled reservations

server/src/repositories/Equipment/ReservationRepository.ts
smellslikekeenspirit: @MatthewGalan https://github.com/rit-construct-makerspace/makerspace/pull/125/files you added creator in the type. Should we get rid of this? I was following this schema

server/src/models/equipment/reservationInput.ts
smellslikekeenspirit: @MatthewGalan you added `creator` and `creatorID` in the types. https://github.com/rit-construct-makerspace/makerspace/pull/125/files I was following this. Should we get rid of this?

MatthewGalan: Good catch. That PR was before I knew about how the context works. Yeah I still think we can remove it.

server/src/resolvers/reservationResolver.ts
```diff
@@ -0,0 +1,51 @@
...
+    createReservation: async (_: any, args: { reservation: ReservationInput }) => {
+      return await reservationRepo.createReservation(args.reservation);
+    },
+
+    assignLabbieToReservation: async (_: any, args: { resId: number, labbieId: number }) => {
+      return await reservationRepo.assignLabbieToReservation(args.resId, args.labbieId);
+    },
+
+    removeLabbieFromReservation: async (_: any, args: { resId: number, labbieId: number }) => {
+      return await reservationRepo.removeLabbieFromReservation(args.resId);
```
MatthewGalan, line 34: After discussion with Mike we decided to get rid of the concept of labbies being assigned to reservations, so I don't think we'll need `assignLabbieToReservation` or `removeLabbieFromReservation` anymore. Likewise we can remove the labbie stuff from the reservation table.

server/src/resolvers/reservationResolver.ts
```diff
@@ -0,0 +1,51 @@
...
+    },
+
+    addComment: async (_: any, args: { resId: number, authorId: number, commentText: string }) => {
```
MatthewGalan, line 37: I don't think we need `authorID` since we can get that from the context.

server/src/repositories/Equipment/ReservationRepository.ts
smellslikekeenspirit: We can, just wanted to see the comment itself lol

server/src/repositories/Equipment/ReservationRepository.ts
```diff
@@ -27,171 +19,104 @@ export class ReservationRepository implements IReservationRepository {
...
-              endTime: reservation.endTime,
-              startingMakerComment: reservation.startingMakerComment
-            },
-            "id"
-          )
-        )[0];
-        return singleReservationToDomain(this.getReservationById(newId));
+  public async noConflicts(reservation: ReservationInput): Promise<boolean> {
+      const conflicts = await this.queryBuilder("Reservations")
+        .select("*")
```
MatthewGalan, line 58: Don't need to include the `"*"`, as `.select()` works the same.

MatthewGalan (approved):
LGTM, only minor comments. Feel free to merge after addressing them.

PR #192, opened 2022-03-27
Archiving Training Modules 

axg3073:
+ small edit to Reservations Archive to return id.

### DIFF
--- a/server/src/repositories/Training/ModuleRepository.ts
+++ b/server/src/repositories/Training/ModuleRepository.ts
@@ -24,7 +24,7 @@ export async function getModules(): Promise<TrainingModule[]> {
   return TrainingModuleMap.trainingModulesToDomain(knexResult);
 }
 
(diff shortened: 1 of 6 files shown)


server/src/repositories/Training/ModuleRepository.ts
```diff
@@ -24,8 +24,9 @@ export async function getModules(): Promise<TrainingModule[]> {
   return TrainingModuleMap.trainingModulesToDomain(knexResult);
 }
 
-export async function deleteModuleById(id: number): Promise<void> {
-  await knex("TrainingModule").where({ id: id }).del();
+export async function archiveModuleById(id: number): Promise<TrainingModule | null> {
```
MatthewGalan, line 27: I think we can rename this to just `archiveModule`, since "by ID" is implied by the argument. This will also make it consistent with `archiveModuleItem`.

MatthewGalan (approved):
LGTM, just one tiny suggestion. Feel free to merge.

PR #194, opened 2022-03-29
Mod sub grading logic implementation

pjw9963:


server/src/resolvers/trainingResolver.ts
```diff
@@ -79,6 +83,70 @@ const TrainingResolvers = {
...
+            if (areEqual) {
+              correct++;
+            } else {
+              incorrect++;
+            }
+          }
+
+          const finalScore = (correct / (correct + incorrect)) * 100;
+
+          const THRESHOLD = 80;
```
MatthewGalan, line 138: I think we should have a constants file for the backend, where we can put this.

server/src/repositories/Training/OptionRepository.ts
```diff
@@ -20,6 +20,17 @@ export async function getOptionsByModuleItem(
   return OptionMap.optionsToDomain(knexResult);
 }
 
+export async function getCorrectOptionsWithModuleItemByModule(
```
MatthewGalan, line 23: I think this name could be improved. My first thought is `getCorrectOptionsForEachModuleItem`, but that's just as long. Conceptually, we're getting all the correct answers. If this was a pen & paper test, that would be called the answer sheet. `getAnswerSheet`?

server/src/resolvers/trainingResolver.ts
```diff
@@ -79,6 +83,70 @@ const TrainingResolvers = {
...
+              submission.moduleID
+            );
+          if (!answers || answers.length === 0) {
+            throw Error(
+              "Training module for provided ID has no correct answers"
+            );
+          }
+          let correct = 0,
+            incorrect = 0;
+          for (let correctAnswers of answers) {
```
MatthewGalan, line 107: Maybe something like this would be more clear?
```suggestion
          for (let moduleItemAnswer of answerSheet) {
```

server/src/schemas/trainingSchema.ts
```diff
@@ -23,6 +23,21 @@ export interface Option {
...
+  selectedOptionIDs: string[];
+}
+
+export interface ModuleSubmissionInput {
+  moduleID: string;
+  userID: string;
+  answers: ModuleItemAnswerInput[];
+}
+
+export interface ModuleItemAnswers {
```
MatthewGalan, line 37: I think it would be easier to understand if we called this `ModuleItemAnswer` instead of plural _answers_. My reasoning is that we frequently have this object as part of a list, so having the singular item have a plural name muddies the waters somewhat.

server/src/resolvers/trainingResolver.ts
```diff
@@ -79,6 +83,70 @@ const TrainingResolvers = {
...
+            let submittedAnswers = submission.answers.find(
+              (x) => x.moduleItemID == correctAnswers.moduleItemID
+            );
+
+            if (!submittedAnswers) {
+              incorrect++;
+              continue;
+            }
+
+            let areEqual =
```
MatthewGalan, line 117: For this variable and the one above, let's prefer `const` over `let` since we're not reassigning them.

MatthewGalan (approved):
All of my changes were opinions & suggestions, so feel free to make what you want of em and merge when you're ready. Overall a nice job. This is a tricky bit of domain logic for sure.

PR #198, opened 2022-03-30
Holds implementation

MatthewGalan:
- Labbies and admins can create, view and remove holds on users from the users page
- Creating & removing holds creates an audit log
- When you have an outstanding hold, a notification appears in the nav bar
- `ifAllowed()` ensures that the user has no outstanding holds before invoking the callback


pjw9963 (approved):
Lots of good work here. I'm guessing we should throw entitynotfound errors instead of returning null from now on

PR #213, opened 2022-04-04
module grading again + little things

pjw9963:
added mod grading, fixed issue with module submission interface not matching db row, and added missing env values to .env.example


server/src/resolvers/trainingResolver.ts
```diff
@@ -69,14 +71,66 @@ const TrainingResolvers = {
     submitModule: async (
       _parent: any,
       args: { moduleID: number; answerSheet: AnswerInput[] },
-      { ifAllowed }: ApolloContext
+      { ifAllowed, user }: ApolloContext
```
MatthewGalan, line 74: I don't think we need `user` here. We should use the one in `ifAllowed`'s callback arguments, since that one is guaranteed to exist.

server/src/resolvers/trainingResolver.ts
```diff
@@ -69,14 +71,66 @@ const TrainingResolvers = {
...
     ) => {
       return ifAllowed(
         [Privilege.ADMIN, Privilege.LABBIE, Privilege.MAKER],
         async (user) => {
-          // TODO: grade answer sheet
-          console.log(args.answerSheet);
-          return null;
+          const module: any = await (
+            await ModuleRepo.getModuleByID(args.moduleID)
+          ).quiz;
```
MatthewGalan, line 81: Would this work?

``` ts
const  { quiz } = await ModuleRepo.getModuleByID(args.moduleID);
```

server/src/resolvers/trainingResolver.ts
```diff
@@ -69,14 +71,66 @@ const TrainingResolvers = {
...
+              return option.correct;
+            });
+
+            let correctOptionIds = correctOptions.map(function (option: any) {
+              return option.id;
+            });
+
+            let submittedItem = args.answerSheet.find((item: any) => {
+              return item.itemID === question.id;
+            });
```
MatthewGalan, line 108: For the `filter`, `map`, and `find` I think it would be cleaner to use arrow functions, like this:

``` ts
const submittedItem = args.answerSheet.find((item: any) => item.itemID === question.id);
```

server/src/resolvers/trainingResolver.ts
```diff
@@ -69,14 +71,66 @@ const TrainingResolvers = {
...
+            incorrect = 0;
+
+          for (let question of module) {
+            if (
+              question.type !== "CHECKBOXES" &&
+              question.type !== "MULTIPLE_CHOICE"
+            )
+              continue;
+
+            let correctOptions = question.options.filter(function (
```
MatthewGalan, line 96: Use `const` over `let` pls

server/src/resolvers/trainingResolver.ts
```diff
@@ -69,14 +71,66 @@ const TrainingResolvers = {
...
+
+            let correctOptionIds = correctOptions.map(function (option: any) {
+              return option.id;
+            });
+
+            let submittedItem = args.answerSheet.find((item: any) => {
+              return item.itemID === question.id;
+            });
+
+            if (submittedItem?.optionIDs === undefined) {
```
MatthewGalan, line 110:
```suggestion
            if (!submittedItem?.optionIDs) {
```

This way it checks for `null` as well.

server/src/resolvers/trainingResolver.ts
```diff
@@ -69,14 +71,66 @@ const TrainingResolvers = {
...
+
+          const grade = (correct / (incorrect + correct)) * 100;
+
+          addTrainingModuleAttemptToUser(
+            user.id,
+            args.moduleID,
+            grade >= MODULE_PASSING_THRESHOLD
+          );
+
+          return grade;
```
MatthewGalan, line 133: Pls add a log as well.

MatthewGalan (changes requested):
Only minor changes requested. Overall this is an awesome implementation! 👏

server/src/resolvers/trainingResolver.ts
pjw9963: yes but I needed to add a type for the `quiz` column in the TrainingModuleRow interface

MatthewGalan: In my example, `quiz` should come back as an `object` type because of https://github.com/rit-construct-makerspace/makerspace/blob/master/server/src/db/tables.ts#L93, no?

The way you have it is fine tbh. Just curious.

MatthewGalan (approved):

PR #215, opened 2022-04-05
removed correct fields from modules for makers

pjw9963:
The `module` and `modules` resolvers now will removed the `correct` field from the json quiz object if the user's privilege is `MAKER`. These resolvers now require the user to be authenticated. I could update the `IsAuthenticated` function to simply return a boolean, and return the stripped quiz view to unauthenticated users as well if you think that's worth it

### DIFF
--- a/server/src/resolvers/trainingResolver.ts
+++ b/server/src/resolvers/trainingResolver.ts
@@ -9,11 +9,46 @@ import { MODULE_PASSING_THRESHOLD } from "../constants";
 
 const TrainingResolvers = {
   Query: {
-    modules: async (_parent: any, args: any, context: any) =>
-      await ModuleRepo.getModules(),
+    modules: async (
+      _parent: any,
+      args: any,
+      { ifAuthenticated }: ApolloContext
+    ) =>
+      ifAuthenticated(async (user) => {
+        let modules = await ModuleRepo.getModules();
 
-    module: async (_parent: any, args: { id: number }, context: any) =>
-      await ModuleRepo.getModuleByID(args.id),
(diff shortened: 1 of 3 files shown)


server/src/resolvers/trainingResolver.ts
```diff
@@ -9,11 +9,46 @@ import { MODULE_PASSING_THRESHOLD } from "../constants";
...
+      ifAuthenticated(async (user) => {
+        let module = await ModuleRepo.getModuleByID(args.id);
+
+        if (user.privilege === "MAKER")
+          for (let item of module.quiz) {
+            if (item.options) {
+              for (let option of item.options) {
+                delete option.correct;
+              }
+            }
```
MatthewGalan, line 47: Can we extract this into a function instead of repeating it for `module` and `modules`?

MatthewGalan (changes requested):
Nice work!

Do we need `ifAuthenticated`? Can't we just use `ifAllowed` and pass in all roles?

pjw9963:
> Nice work!
> 
> Do we need `ifAuthenticated`? Can't we just use `ifAllowed` and pass in all roles?

Ifallowed will throw an error if the user has any holds, do we want to block a user viewing modules due to a hold?

MatthewGalan:
> > Nice work!
> > Do we need `ifAuthenticated`? Can't we just use `ifAllowed` and pass in all roles?
> 
> Ifallowed will throw an error if the user has any holds, do we want to block a user viewing modules due to a hold?

Hmm... good point. LGTM then.

PR #223, opened 2022-04-07
Delete user from Front End

axg3073:
-Delete user from front-end
-Check privileges levels for deletion
-Log it


client/src/pages/admin/users/UserModal.tsx
```diff
@@ -101,6 +110,26 @@ export default function UserModal({ selectedUserID, onClose }: UserModalProps) {
     });
   };
 
+  const handleDeleteUserClicked = () => {
+      let fName = getUserResult.data.user.firstName;
+      let lName = getUserResult.data.user.lastName;
+      let privilege = getUserResult.data.user.privilege;
+      if (privilege == Privilege.ADMIN) {
```
MatthewGalan, line 117: Let's just hide the delete user button from labbies, that way we don't have to do a privilege check here.

client/src/pages/admin/users/UserModal.tsx
```diff
@@ -101,6 +110,26 @@ export default function UserModal({ selectedUserID, onClose }: UserModalProps) {
     });
   };
 
+  const handleDeleteUserClicked = () => {
+      let fName = getUserResult.data.user.firstName;
+      let lName = getUserResult.data.user.lastName;
+      let privilege = getUserResult.data.user.privilege;
```
MatthewGalan, line 116: Let's use `const` instead of `let` whenever we're not reassigning the variable. Same goes for `result` below.

client/src/pages/admin/users/UserModal.tsx
```diff
@@ -101,6 +110,26 @@ export default function UserModal({ selectedUserID, onClose }: UserModalProps) {
...
+              deleteUser({
+                  variables: {userID: getUserResult.data.user.id},
+                  refetchQueries: [{query: GET_USER, variables: {id: selectedUserID}}],
+              });
+          };
+      }
+      else {
+          window.alert("You don't have the privilege level to delete users!")
+      }
+  }
```
MatthewGalan, line 131: Are you running Prettier? This file should be formatted with 2 spaces automatically for you. Perhaps we can look into it together in our next meeting.

MatthewGalan (changes requested):

server/src/resolvers/usersResolver.ts
```diff
@@ -77,12 +77,25 @@ const UsersResolvers = {
...
-      context: any
-    ) => {
-      return await UserRepo.archiveUser(args.userID);
+      {ifAllowed}: ApolloContext) => {
+
+      return ifAllowed(
+        [Privilege.ADMIN],
+        async (user) => {
+
+          const userSubject = await UserRepo.getUserByID(args.userID);
```
MatthewGalan, line 89: Make sure you're running Prettier on the backend too pls

client/src/pages/admin/users/UserModal.tsx
```diff
@@ -171,18 +194,17 @@ export default function UserModal({ selectedUserID, onClose }: UserModalProps) {
...
-                startIcon={<DeleteIcon />}
-                onClick={() =>
-                  window.confirm(
-                    "Are you sure you wish to delete John Smith's account? This cannot be undone."
-                  )
-                }
-              >
-                Delete account
-              </Button>
+              {getUserResult.data.user.privilege == Privilege.ADMIN &&
```
MatthewGalan, line 197: It's generally better to use `===` instead of `==`, since `==` does weird type conversion stuff.

MatthewGalan (approved):
Ready to merge after addressing final feedback

PR #224, opened 2022-04-08
added Exp. Date for Submitted Modules

axg3073:


server/src/db/migrations/20220324210801_module_submissions_table.ts
```diff
@@ -11,6 +11,7 @@ export async function up(knex: Knex): Promise<void> {
           t.integer('makerID').references('id').inTable('Users');
           t.timestamp("submissionDate").defaultTo(knex.fn.now());
           t.boolean("passed");
+          t.timestamp("expirationDate").defaultTo(knex.raw(`? + ?::INTERVAL`, [knex.fn.now(), '365 day']));
```
MatthewGalan, line 14: Can you create a new migration instead of modifying existing ones pls. Now that we have a DB deployed on AWS, editing existing migrations will break things.

MatthewGalan (changes requested):

MatthewGalan (approved):

PR #228, opened 2022-04-10
Hashing UIDs with salt

axg3073:


server/src/resolvers/usersResolver.ts
```diff
@@ -52,11 +56,18 @@ const UsersResolvers = {
       },
       context: any
     ) => {
-      const hashedUniversityID = hashUniversityID(args.universityID);
+
+      const length = 16; //salt length
+
+      const salt = randomBytes(Math.ceil(length/2))
```
MatthewGalan, line 62: Why are we doing `Math.ceil(length/2)`? Why not just put do `randomBytes(16)`?

server/src/resolvers/usersResolver.ts
```diff
@@ -52,11 +56,18 @@ const UsersResolvers = {
...
+        .toString('hex')
+        .slice(0,length);
```
MatthewGalan, line 64: Why are we calling `slice`?

server/src/resolvers/roomsResolver.ts
```diff
@@ -63,8 +63,18 @@ const RoomResolvers = {
       const room = await RoomRepo.getRoomByID(args.roomID);
       assert(room);
 
-      const hashedUniversityID = hashUniversityID(args.universityID);
-      const user = await UserRepo.getUserByUniversityID(hashedUniversityID);
+      let hashedUIDString = "";
+
+      const users = await UserRepo.getUsers();
+      users.forEach(user => {
```
MatthewGalan, line 69: Why are we looping through every user in the system? Can't we just grab the hashed UID from the current user using the callback argument in `ifAllowed`?

server/src/resolvers/usersResolver.ts
```diff
@@ -3,16 +3,20 @@ import * as ModuleRepo from "../repositories/Training/ModuleRepository";
...
 }
 
-export function hashUniversityID(universityID: string) {
-  return createHash("sha256").update(universityID).digest("hex");
+export function hashUniversityID(universityID: string, salt: string) {
+  const hashed = createHash("sha256").update(universityID + salt).digest("hex");
+  return {
+    salt: salt,
+    hashedValue: hashed
+  };
```
MatthewGalan, line 19: What's the benefit of returning the hash and salt separately, as an object? I think if we just returned `hashed + ":" + salt`, then we wouldn't have to stitch them together every time we want to check a UID.

MatthewGalan (changes requested):

server/src/resolvers/roomsResolver.ts
axg3073: will the current User always be the swiped in user? If so, would makes sense with that route

PR #233, opened 2022-10-18
Modal exits

mattcorrea8:


jrc1697 (approved):
Looks good, will be reusable for future modal components

PR #235, opened 2022-10-19
Training completion animation

mattcorrea8:


jrc1697 (approved):
Fixed issue with deletes not persisting when route changes. Looks good, should be reusable for future notifications in the frontend

PR #239, opened 2022-10-27
Maker training animations and cancel quiz button

mattcorrea8:


jrc1697 (changes requested):
Animations look good.
Would it be possible to warn them about unsaved progress when they try to navigate away from the page? maybe something like [this:](https://stackoverflow.com/a/45869459/14489899)

PR #243, opened 2022-11-14
Fixed handling of canceling the placing of a hold

mattcorrea8:
Before:
When an admin places a hold, a prompt appears to describe the reason for the hold. If the admin were to press cancel, a message would appear saying the description cannot be empty. Same handling for submitting an empty description.

Now: 
If the admin presses cancel, the window will close regardless of description contents. When admin presses submit on an empty description, it shows message that description cannot be empty and does not submit the hold.


jrc1697 (approved):
Looks good, the fix works

PR #244, opened 2022-11-14
Added question duplication functionality

mattcorrea8:
Ability to duplicate training questions added


jrc1697 (approved):
Looks good to me, code is very reusable as well.
Maybe we could consider adding the duplicated item next to the original item at some point, but this meets the MVP.

PR #246, opened 2022-11-15
updated user terminology both front and back end

axs9956:


jrc1697 (approved):
Staff and Mentor were switched in a few places, so I fixed that. Also fixed the bug where it used the privilege of the user being viewed on the frontend, instead of the current user.

PR #247, opened 2022-11-25
Empty Training Module's Should Not Be Saved

mattcorrea8:
Can no longer save a training module if it contains no items. Any pre-existing module whose questions are all deleted will be reverted back to the saved state prior to edit.


jrc1697:
I'm thinking this message might make more sense in a MUI Alert (maybe it could pop up at the top of the screen w/ absolute positioning?). Just because it's an error not a successful action like the other toast messages were.

jrc1697:
We may want to enforce this on the backend as well

PR #252, opened 2023-02-05
Material and equipment search bugs

mattcorrea8:


jrc1697 (approved):
Looks good, search is case insensitive

PR #263, opened 2023-11-08
Updating SAML specifications

lollygagger:
updated identifier format to match ITS requested format on heroku deployment


CharltonL (approved):
👍

PR #493, opened 2025-05-13
Hisotry Time HH:MM:SS

### LINKED ISSUE #487: History Show Session Time in HH:MM:SS
Session time is currently reported in seconds. Saving it this way is a good idea for the backend, but on the history it would be nice if it was shown in seconds, minutes and seconds, or hours, minutes, and seconds depending on session length.

ZZwatermelon:
Closes #487

@evstod could you test this when you get a chance? I'm not setup to send fake calls to my own device yet.


ZZwatermelon:
based on
https://stackoverflow.com/questions/1322732/convert-seconds-to-hh-mm-ss-with-javascript

PR #516, opened 2025-05-17
Fixing the CI

ZZwatermelon:
A PR so I can watch the CI run


ZZwatermelon:
The problem with the frontend CI was react-markdown required react ^18.0.0 while most other packages (such as material-ui) require react ^17.0.0

PR #731, opened 2025-07-29
Hours Overhaul

ZZwatermelon:

PR #733, opened 2025-07-30
Hours and Private Submodule

cowsed:
- #731 New Hour storage format and UI 
- #732 Use a private git submodule to keep Atrium API code private


.gitmodules
ZZwatermelon, line 1: @cowsed can you make this submodule more clickable

cowsed: no

ZZwatermelon (approved):

PR #743, opened 2025-07-31
Papercut integration with Construct Credits

cowsed:
add FREE_3D_PRINTS environment variabel

### DIFF
--- a/client/src/common/CurrentUserProvider.tsx
+++ b/client/src/common/CurrentUserProvider.tsx
@@ -16,7 +16,6 @@ export const GET_CURRENT_USER = gql`
       lastName
       privilege
       setupComplete
-      balance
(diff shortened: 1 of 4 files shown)


client/src/common/CurrentUserProvider.tsx
```diff
@@ -38,6 +37,10 @@ export const GET_CURRENT_USER = gql`
       manager
       staff
       trainer
+      currencyAccount{
+        id
+        balance
+      }
```
ZZwatermelon, line 43: :fire:

server/src/integrations/currency/currency.ts
ZZwatermelon, line 1: got some ocnsole logs here

PR #744, opened 2025-07-31
Zone status hotfix

ZZwatermelon:


cowsed (approved):
really padding out the githuib stats with this one

PR #751, opened 2025-08-01
Node Package Updates and Security Fixes (#750)

cowsed:
Apollo server to v5
Express to v5
Removed a bunch of unused dependencies (what was google api doing in there) Removed dependencies for nonexistent test framework Updated Passport SAML to mitigate bypassing SAML
maybe some other things
maybe broke shibbing due to breaking changes in Passport SAML


ZZwatermelon (approved):
security shmecurity, we should just delete the DB so there's nothing for anyone to compromise

PR #778, opened 2025-08-06
Archive Users

### LINKED ISSUE #481: Remove Graduated Students
We currently have no mechanism to remove accounts of students who have graduated. We need to figure something out. 

Re-investigating getting the student account state from CLAWS would be ideal, but that'll be a challenge I fear. 

Alternatively, we gather the expected graduation date when students register. If we soft-lock accounts who've past this date, that can act as a good catch for 95% of cases. Students who lied would still have access, though. 

If we go that route, we also need a mechanism to fix users who have an incorrect graduation date (set too soon for whatever reason).

### LINKED ISSUE #579: Permissions & Roles Overhaul
Implement a new per-makerspace permission system as described in the image below.

Image

ZZwatermelon:
Archive users if they do not have roles defined in the whitelist.
Adds an override on their users page in the case of error.

Closes #481, and due to that also closes #579

### DIFF
--- a/client/src/pages/lab_management/users/UserModal.tsx
+++ b/client/src/pages/lab_management/users/UserModal.tsx
@@ -28,6 +28,7 @@ import LockIcon from '@mui/icons-material/Lock';
 import BlockIcon from '@mui/icons-material/Block';
 import DeleteIcon from '@mui/icons-material/Delete';
 import { LoadingButton } from "@mui/lab";
+import ManageUserArchive from "./ManageUserArchive";
 
 const StyledInfo = styled.div`
   margin-top: 16px;
@@ -100,6 +101,8 @@ export const GET_USER = gql`
       ritUsername
       cardTagID
       notes
+      isArchived
+      forceArchive
       holds {
         id
         creator {
@@ -218,7 +221,6 @@ const DELETE_PASSED_MODULE = gql`
     deletePassedModule(userID: $userID, moduleID: $moduleID)
   }
 `;
-
 
 interface UserModalProps {
   selectedUserID: string;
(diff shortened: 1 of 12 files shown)


client/src/pages/lab_management/users/UserModal.tsx
```diff
@@ -24,6 +24,15 @@ import { FullZone, GET_FULL_ZONES } from "../../../queries/zoneQueries";
 import LockIcon from '@mui/icons-material/Lock';
 import BlockIcon from '@mui/icons-material/Block';
 import DeleteIcon from '@mui/icons-material/Delete';
+import { LoadingButton } from "@mui/lab";
+import ManageUserArchive from "./ManageUserArchive";
+
+const StyledInfo = styled.div`
+  margin-top: 16px;
```
cowsed, line 31: what the heck cant believe you're bringing back unused variables

cowsed: i think loading button is also unused

ZZwatermelon: I'm sorry, I wasn't sure from looking at the merge conflict resolver :(

cowsed:
> > this is i think just paranoia but does the nullable boolean get passed through graphql correctly when you're unforcing something. like the were force allowed but then you change that to be unforced and it sends a null through
> 
> It does, I tested it

big

server/src/auth.ts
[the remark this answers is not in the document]
ZZwatermelon: shhhh

client/src/top_nav/TopNav.tsx
```diff
@@ -54,6 +54,13 @@ export default function TopNav() {
 
     return (
       <Stack>
+        { // Archived account alert
+          currentUser.archived
+            ? <Alert variant="filled" severity="error" sx={{ borderRadius: 0 }}>
+              Your account has been archived. You are still able to take trainings and review reference materials, but you won't be able to use any machines. Please speak to a member of staff if you believe this was a mistake.
+            </Alert>
```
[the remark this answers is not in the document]
ZZwatermelon: I mostly didn't want to add that because its already one of the longest alert messages. I feel like if they can't come in person to speak with staff, being locked out of the machines won't effect them?

cowsed (approved):

PR #780, opened 2025-08-06
Fix assorted deprecations and unused packages in client

cowsed:
- Moved react-scripts to be a devDependency (since it is). running `npm audit --production` shows 0 vulnerabilities now. react-scripts will not be used in the actual server (for now it will be there but not be called bc heroku runs in staging)
-


package.json
```diff
@@ -80,9 +80,7 @@
     "@types/node-schedule": "^2.1.7",
     "@types/passport": "^1.0.11",
     "@types/passport-local": "^1.0.35",
-    "@types/passport-saml": "^1.1.3",
```
cowsed, line 83: this is still used, but it shouldnt be in both dependencies and dev dependencies (same with a lot of these but this is the only one i tracked down)

ZZwatermelon (approved):
seems groovy

PR #809, opened 2025-08-08
No More REST ACS API

cowsed:
All shlugs and welcome readers are on WSACS (websocket access control system)


cowsed:
that was really mean of you

PR #822, opened 2025-08-12
Remove uses/db columns that were only required for WS ACS API

### LINKED ISSUE #537: Actions when Shlug Status Rest API is Deprecated  (non-imminent)
When the Shlug Rest API is deprecated and all auths, states go through websocket, we can cleanup the database a bit
Once this is the case, EquipmentInstance will hold a foreign key to the Reader
The equipment/equipmentInstance will hold its equipment type, makerspace, room, etc
so, the machineId, machineType, zone can be removed from the Readers table

name will stick around since we will use this to give a human readable name to the device

cowsed:
Closes #537

### DIFF
--- a/server/src/wsapi.ts
+++ b/server/src/wsapi.ts
@@ -263,10 +263,6 @@ async function authorizeUIDToUnlock(uid: string, readerId: number, inResponse: S
     if (machineInst) {
       try {
         machine = await getEquipmentByID(machineInst.equipmentID);
-        if (machine == undefined) {
-          // bizzare error handling bc api getters can be inconsistent
-          throw EntityNotFound;
-        }
       } catch (EntityNotFound) {
         machine = undefined;
       }


client/src/pages/lab_management/manage_equipment/EquipmentEditor.tsx
```diff
@@ -142,6 +142,8 @@ export default function EquipmentEditor({
     window.location.reload()
   }
 
+  console.log(equipment.trainingModules.length);
```
ZZwatermelon: :raised_eyebrow:

client/src/pages/lab_management/readers/ReaderCard.tsx
```diff
@@ -54,7 +54,7 @@ export default function ReaderCard({ reader, makerspaceID, searchQuery }: Reader
   const theme = useTheme();
 
   const machineResult = useQuery(GET_CORRESPONDING_MACHINE_BY_READER_ID_OR_MACHINE_ID, {
```
ZZwatermelon: Should this query get renamed?

server/src/db/migrations/20250812144441_cleanup_readers_table.ts
```diff
@@ -0,0 +1,27 @@
+import type { Knex } from "knex";
+
+
+export async function up(knex: Knex): Promise<void> {
+    if (! await knex.schema.hasColumn("Readers", "machineType")) {
```
ZZwatermelon, line 5: Don't you want this the other way (i.e. don't negate it)
because if it still has the column, you want the migration to run

server/src/db/migrations/20250812144441_cleanup_readers_table.ts
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    await knex.schema.alterTable("Readers", (t)=>{
+        t.dropColumns("machineType", "machineID", "zone", "helpRequested");
+    })
+}
+
+
+export async function down(knex: Knex): Promise<void> {
+    if (await knex.schema.hasColumn("Readers", "machineType")) {
```
ZZwatermelon, line 16: same as above, flip logic

server/src/resolvers/readersResolver.ts
```diff
@@ -166,16 +166,13 @@ const ReadersResolver = {
   Mutation: {
     /**
      * Create a Reader
-     * @argument machineID ID of Equipment
-     * @argument machineType Type indication string (mostly deprecated)
      * @argument name Reader name
-     * @argument zone comma seperated list of zone IDs the machine resides in (usually just the one)
      * @returns new Reader
      * @throws GraphQLError if not STAFF or is on hold
```
ZZwatermelon, line 171: Actually throws if not MANAGER or higher

server/src/wsapi.ts
```diff
@@ -284,7 +284,7 @@ async function authorizeUIDToUnlock(uid: string, readerId: number, inResponse: S
     // Find Machine
     if (machine == null) {
       if (reader?.SN == null) {
-        wsApiLog("{user} failed to swipe into a machine with error '{error}'", "auth", { id: user.id, label: getUsersFullName(user) }, { id: reader.machineID, label: `Machine ${reader.machineID} does not exist` });
+        wsApiLog("{user} failed to swipe into a machine with error '{error}'", "auth", { id: user.id, label: getUsersFullName(user) }, { id: 0, label: `Machine does not exist` });
```
ZZwatermelon, line 287: This error could be better, it seems like it only happens if the shlug doesn't exist in the DB, which feels like it should be a separate check from if the machine exists

server/src/wsapi.ts
```diff
@@ -284,7 +284,7 @@ async function authorizeUIDToUnlock(uid: string, readerId: number, inResponse: S
     // Find Machine
     if (machine == null) {
```
ZZwatermelon, line 285: machine: EquipmentRow | undefined
not null

maybe declare it as undefined on Line 262 to be more explicit?
The whole process of getting the equipment around line 262-273 feels very roundabout and like it could be improved, but maybe out of scope of this PR?

server/src/db/migrations/20250812144441_cleanup_readers_table.ts
```diff
@@ -0,0 +1,30 @@
...
+
+
+export async function up(knex: Knex): Promise<void> {
+    if (! await knex.schema.hasColumn("Readers", "machineType")) {
+        return;
+    }
+
+    await knex.schema.alterTable("Readers", (t)=>{
+        t.dropColumns("machineType", "machineID", "zone", "helpRequested");
+        // t.("SN");
```
ZZwatermelon, line 11: :raised_eyebrow:

server/src/wsapi.ts
```diff
@@ -259,11 +259,11 @@ async function authorizeUIDToUnlock(uid: string, readerId: number, inResponse: S
     // Find Machine Instance
     const machineInst = await getInstanceByReaderID(readerId);
 
-    var machine: EquipmentRow | undefined;
+    var machine: EquipmentRow | undefined = undefined;
     if (machineInst) {
       try {
         machine = await getEquipmentByID(machineInst.equipmentID);
-        if (machine == null) {
+        if (machine == undefined) {
```
ZZwatermelon, line 266: === ?

Also I don't think it is possible for that function to return undefined?
And if it does return undefined we're just setting it to undefined anyway, so I don't think this check is necessary

server/src/db/migrations/20250812144441_cleanup_readers_table.ts
cowsed: :shushing_face:

ZZwatermelon (approved):
:fireworks: :finland: :fireworks:

PR #839, opened 2025-08-14
Dev SAML

### LINKED ISSUE #834: Fix the dev auth so it properly serializes / deserializes users
Currently we don't really use dev auth, and that leads to weird behaviors that make it hard to test things affecting auth

ZZwatermelon:
Closes #834


server/src/auth.ts
```diff
@@ -41,36 +42,64 @@ interface RitSsoUser {
...
-      lastName: devUser.lastName,
-      ritUsername: devUser.ritUsername
-    };
-  }
-}
-
 // Map the test users from samltest.id to match
 // the format that RIT SSO will give us.
 function mapSamlTestToRit(testUser: any): RitSsoUser {
-  console.log("MAP TEST USER: " + testUser["urn:oid:2.5.4.42"]);
```
cowsed, line 66: do u always wanna print it

server/src/auth.ts
```diff
@@ -88,85 +117,168 @@ export function setupSessions(app: express.Application) {
...
 
+  app.post("/logout", (req, res) => {
     if (req.session) {
       req.session.destroy((err) => {
         if (err) {
           res.status(400).send("Logout failed");
         } else {
-          res.clearCookie("connect.sid");
-
+          // res.clearCookie("connect.sid");
```
cowsed, line 273: should this be here

cowsed (approved):
siurely this wont break make login

server/src/auth.ts
cowsed: i admit defeat

cowsed (approved):

PR #841, opened 2025-08-14
sTORE Quick Fixes

cowsed:


client/src/pages/lab_management/inventory/MaterialModalContents.tsx
```diff
@@ -221,6 +224,16 @@ export default function MaterialModalContents({
         onChange={handleNotesChanged}>
       </TextareaAutosize>
 
+      <TextareaAutosize
```
ZZwatermelon, line 227: :vomiting_face: me when we're not using TextField

client/src/pages/lab_management/storefront/ListingCard.tsx
```diff
@@ -14,7 +14,7 @@ interface ListingCardProps {
 export function ListingCard(props: ListingCardProps) {
   const currentUser = useCurrentUser();
   const isMobile = useIsMobile();
-
+  console.log(props.item.image)
```
ZZwatermelon, line 17: prolly don't want to log this

client/src/pages/lab_management/inventory/MaterialModalContents.tsx
ZZwatermelon: (this is non-blocking, to be clear)

ZZwatermelon (changes requested):
just the errant log statement

PR #847, opened 2025-08-15
No more zonehours/OpenHours

cowsed:
Remove old zone hours stuff


ZZwatermelon (approved):
:falkland_islands: down with the king

PR #865, opened 2025-08-17
Stateful Themes

### LINKED ISSUE #795: Store light and dark themes in state
Storing the light and dark themes in state would allow dynamic switching without having to reload the page.

ZZwatermelon:
Closes #795 
definitely no secret olive garden theme, why would there be a secret olive garden theme, that would be crazy


cowsed (approved):
nice

PR #872, opened 2025-08-19
Replaced logos

JimSHED:
Turns out we are called "SHED" not "The SHED". Changed out the logos with the same file name so hopefully it just works?


ZZwatermelon (changes requested):
The logos being shorter affect how they are displayed on the page.
The end result is that the top navbar ends up being much taller than it was before.

image

ZZwatermelon:
IMO we should really merge in #865 before we fix the logo height issue, as that PR makes it much easier to edit the topnav bar (and if we do it the other way I fear the merge conflict will be more ugly)

PR #925, opened 2025-09-03
Transaction Processing

### LINKED ISSUE #912: Incremental Transactions
Currently, each transaction generates a receipt. But, there are many things we do that actually result in multiple successive transactions all related to the same thing. We should capture these all into a relevant transaction history as one blob. 

Ledger should remain a record of every time we touch money, but this makes what the transactions mean more understandable. 

Example: Submitting a resin print results in a $0.01 "Job Created" transaction, then a $XX.YY "Quoted with actual price" transaction. Then can be followed by a -$4.20 "Refund for partial failure". And that is all captured under the umbrella of Job #12345.

### LINKED ISSUE #913: Do not Immediately charge <$0.03 first charge from 3DPrinterOS
3DPrinterOS has a global setting to enable transactions on non-GCODE 3D prints (i.e. STLs submitted for full-service). So to be able to charge for some full-serve 3D prints but not others, every print needs a $0.01 or so cost associated. So if the print is less than $0.03, we should not charge that amount to CC/Tiger Bucks. 

BUT, 3DPrinterOS only allows us to adjust the price, not set a new price, so if we have the placeholder $0.01, quote the job at $12.00, it will charge the user $11.99. So, we should hold onto that 1 cent, and add it to the price when that gets charged. 

It is a gross solution, but it will fix things.

### LINKED ISSUE #914: Receipt should show currency breakdown
A receipt should show how much of the charge came out of or went in to CC and Tiger Bucks, respectively.

cowsed:
Closes #912
Closes #913
Closes #914


server/src/integrations/currency/currency.ts
```diff
@@ -72,144 +37,131 @@ export async function getAccountBalance(username: string): Promise<number | Make
...
 
-    // Refunds should usually be handled by reversePreviousTransaction. But, if we are trying to charge a negative amount, we don't know if it should go to TB or CC. To be safe, send it to CC and we can manually resolve it
-    if (transaction.isRefund()) {
-      const success = await CurrencyAccountRepo.adjustAccountBalanceIfAvailableCents(makeAccountID, deltaCents, transaction.source, transaction.description, printerJobId, true);
-      return success;
-    }
+  const availToRefund = split.atrium + split.credit;
+  if (cents > availToRefund){
+    return MakeMoneyError.RefundTooLarge;
+  }
```
ZZwatermelon, line 130: I'm not sure what this is doing, I think the comment on this function could do a better job explaining what the cents variable is for.

server/src/integrations/currency/currency.ts
```diff
@@ -72,144 +37,131 @@ export async function getAccountBalance(username: string): Promise<number | Make
...
 
-    const remainingToCharge = await CurrencyAccountRepo.chargeAccountReturnRemainingCents(makeAccountID, transaction.subtotal(), transaction.source, transaction.description, printerJobId, false);
-    if (remainingToCharge == 0) {
-      // all was taken care of with CC, don't need to charge atrium
-      return true;
-    }
+  let toAtrium = split.atrium;
+  if (toAtrium > cents) {
+    toAtrium = cents;
+  }
```
ZZwatermelon, line 136: Same as the other cents comment

server/src/integrations/currency/currency.ts
```diff
@@ -72,144 +37,131 @@ export async function getAccountBalance(username: string): Promise<number | Make
...
-    return false;
-  } catch (e) {
-    console.error(`Failed to adjust balances for ${username} - ${JSON.stringify(transaction)}: ${e}`);
-    return false;
   }
-}
+  
+  let toCredit = cents - toAtrium;
+  if (toCredit > split.credit) {
+    console.warn(`Currency: At some point keeping track of currency failed. Trying to refund ${toCredit} but only ${split.credit} available`);
```
ZZwatermelon, line 152: I think I get this a little better now? Basically whenever a currency happens, we keep track of where it came from / where it went to. And so if somewhere along the way we mess it up we could end up in a situation where the numbers don't add up, and in that case we just give the difference in CC to avoid creating money.

I still feel like this could be commented better b/c it is a little confusing.

server/src/integrations/currency/transactions.ts
```diff
@@ -0,0 +1,99 @@
...
+ */
+
+import { CurrencySource, MakeMoneyError } from "./types.js"
+import * as TransactionRepo from "../../repositories/Currency/TransactionRepository.js"
+import * as Currency from "./currency.js"
+import { send_transaction_email } from "../email/email.js";
+
+function isOutstandingCharge(cents: number) {
+    return cents >= -3 && cents < 0;
+}
```
ZZwatermelon, line 13: A description of what an outstanding charge is / why this is like this would be good

server/src/integrations/currency/transactions.ts
```diff
@@ -0,0 +1,99 @@
...
+
+export async function NewTransaction(accountId: number, initialDeltaCents: number, source: CurrencySource, description: { text: string, data: unknown }, options: { printerJobId: number }): Promise<number | MakeMoneyError> {
+    let outstanding = 0;
+    if (isOutstandingCharge(initialDeltaCents)) {
+        // this is an outstanding one, dont charge yet
+        outstanding = initialDeltaCents;
+    }
+    const tid = await TransactionRepo.createTransaction(accountId, source, description, outstanding, options.printerJobId);
+    if (tid == undefined) {
```
ZZwatermelon, line 22: === ?

server/src/integrations/currency/transactions.ts
```diff
@@ -0,0 +1,99 @@
...
+ * 
+ * @param transactionID 
+ * @param deltaCents 
+ * @param reason 
+ * @returns undefined if the amounts were charged
+ */
+export async function UpdateTransaction(transactionID: number, deltaCents: number, reason: string): Promise<MakeMoneyError | { atrium: number, credit: number }> {
+    // check if there is outstanding charges
+    const parent = await TransactionRepo.getTransactionById(transactionID);
+    if (parent == null) {
```
ZZwatermelon, line 47: returns undefined, not null
=== ?

server/src/integrations/currency/transactions.ts
```diff
@@ -0,0 +1,99 @@
...
+    if (entryId == null) {
+        console.error("Currency: Failed to create transaction ID");
+        return MakeMoneyError.SomethingElse;
+    }
+    let amounts = { atrium: 0, credit: 0 };
+
+    if (centsToCharge > 0) {
+        if (split == null) {
+            const res = await Currency.refundCreditAccount(parent.accountID, centsToCharge, parent.origin, reason, entryId)
+            if (typeof res == "boolean" && res == true) {
```
ZZwatermelon, line 67: === ?

server/src/integrations/currency/transactions.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    const entryId = await TransactionRepo.createTransactionUpdate(parent.id, centsToCharge, reason)
+    if (entryId == null) {
+        console.error("Currency: Failed to create transaction ID");
+        return MakeMoneyError.SomethingElse;
+    }
+    let amounts = { atrium: 0, credit: 0 };
+
+    if (centsToCharge > 0) {
+        if (split == null) {
```
ZZwatermelon, line 65: function returns undefined, not null
=== ?

server/src/integrations/currency/transactions.ts
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    let centsToCharge = deltaCents;
+    if (parent.outstandingCharge) {
+        centsToCharge += parent.outstandingCharge;
+    }
+    let split = await TransactionRepo.getChargeSplitForTransactionById(transactionID);
+
+    const entryId = await TransactionRepo.createTransactionUpdate(parent.id, centsToCharge, reason)
+    if (entryId == null) {
```
ZZwatermelon, line 58: returns undefined, not null
=== ?

server/src/integrations/currency/transactions.ts
```diff
@@ -0,0 +1,99 @@
...
+        return MakeMoneyError.SomethingElse;
+    }
+    let amounts = { atrium: 0, credit: 0 };
+
+    if (centsToCharge > 0) {
+        if (split == null) {
+            const res = await Currency.refundCreditAccount(parent.accountID, centsToCharge, parent.origin, reason, entryId)
+            if (typeof res == "boolean" && res == true) {
+                amounts = { atrium: 0, credit: centsToCharge };
+            } else if (typeof res == "string") {
```
ZZwatermelon, line 69: === ?

server/src/integrations/currency/transactions.ts
```diff
@@ -0,0 +1,99 @@
...
+            if (typeof res == "string") {
+                return res;
+            } else {
+                amounts = res;
+            }
+        }
+    } else {
+        const amt = Math.abs(centsToCharge);
+        const res = await Currency.chargeAccount(parent.accountID, amt, parent.origin, reason, entryId);
+        if (typeof res == "string") {
```
ZZwatermelon, line 88: === ?

server/src/integrations/currency/types.ts
ZZwatermelon, line 1: nitpick, capitalization is different for the MakeMoneyError

server/src/integrations/papercut/papercut.ts
```diff
@@ -133,13 +168,69 @@ function printCommentParser(comment: string): { operation: "new" | "failed" | "c
...
+}
+
+async function process3dPrintTransaction(username: string, amount: number, transaction: PrinterTransaction): Promise<MakeMoneyError | boolean> {
+  const account = await getAccountIDByUsername(username);
+  if (account == null) {
+    console.error("no account here");
+    return MakeMoneyError.NoAccount;
+  }
+  const existing = await getTransactionByPrinterJobId(transaction.jobID);
+  if (transaction.type == PrinterTransactionType.New) {
```
ZZwatermelon, line 194: === ?

server/src/integrations/papercut/papercut.ts
```diff
@@ -133,13 +168,69 @@ function printCommentParser(comment: string): { operation: "new" | "failed" | "c
...
+    case "Quote Updated Price":
+      return { type: PrinterTransactionType.QuoteUpdated, jobID };
+    default:
+      return { type: PrinterTransactionType.Other, jobID, customMessage: res[1] };
+  }
+}
+
+async function process3dPrintTransaction(username: string, amount: number, transaction: PrinterTransaction): Promise<MakeMoneyError | boolean> {
+  const account = await getAccountIDByUsername(username);
+  if (account == null) {
```
ZZwatermelon, line 189: returns undefined, not null
=== ?

server/src/integrations/papercut/papercut.ts
```diff
@@ -133,13 +168,69 @@ function printCommentParser(comment: string): { operation: "new" | "failed" | "c
...
+
+    const res = await NewTransaction(
+      account,
+      amount,
+      CurrencySource.Printers,
+      { text: `New 3D Printer Job: ${transaction.jobID}`, data: {} },
+      { printerJobId: transaction.jobID }
+    )
+    if (typeof res == "string") {
+      if (res == MakeMoneyError.NoFunds) {
```
ZZwatermelon, line 207: === ?

server/src/integrations/papercut/papercut.ts
```diff
@@ -133,13 +168,69 @@ function printCommentParser(comment: string): { operation: "new" | "failed" | "c
...
+      } else {
+        return res;
+      }
+    } else {
+      return true;
+    }
+  }
+
+  // Update
+  if (existing == undefined) {
```
ZZwatermelon, line 218: === ?

server/src/integrations/papercut/papercut.ts
```diff
@@ -133,13 +168,69 @@ function printCommentParser(comment: string): { operation: "new" | "failed" | "c
...
+    console.error(`3DPrinterOS: Ignoring money charge for job id: ${transaction.jobID}. Couldn't find transaction for it`);
+    return false;
   }
-  return undefined;
+  console.info("updating");
+  const res = await UpdateTransaction(existing.id, amount, `${typeToString(transaction.type) + (transaction?.customMessage ? (" - " + transaction?.customMessage) : "")}`)
+  if (typeof res == "string") {
+    if (res == MakeMoneyError.NoFunds) {
```
ZZwatermelon, line 225: === ?

server/src/integrations/papercut/papercut.ts
```diff
@@ -176,68 +267,22 @@ async function papercut_adjustUserAccountBalanceIfAvailable(res: any, params: XM
...
-        } else {
-          // multiple transactions have already happened with this print id, possible multiple refund?
-          // figure this out manually to avoid infinite money glitch
-        }
-      }
-    } else {
-      xmlrpcRespondFault(res, 400, "Could not parse print comment");
-      console.error("Could not parse print comment to get job id. Did 3d printer os change their format?", params);
+    const transaction = printCommentParser(comment)
+    if (transaction == null) {
```
ZZwatermelon, line 272: returns undefined, not null
=== ?

server/src/integrations/papercut/papercut.ts
```diff
@@ -176,68 +267,22 @@ async function papercut_adjustUserAccountBalanceIfAvailable(res: any, params: XM
...
-
-      if (printDetails && printDetails.operation == "new") {
-        subject = `3D Print Job #${printDetails.jobID}`
-      } else if (printDetails && (printDetails.operation == "cancelled" || printDetails.operation == "failed")) {
-        subject = `3D Print Refund Job ${printDetails.jobID}`;
-      }
-      await send_transaction_email(username + "@rit.edu", subject, transaction, constructCreditsAfter);
+    const amountCents = Math.round(adjustment * 100);
+    const result = await process3dPrintTransaction(username, amountCents, transaction);
+    if (typeof result == "string") {
```
ZZwatermelon, line 278: === ?

ZZwatermelon (changes requested):

server/src/db/tables.ts
```diff
@@ -554,20 +555,73 @@ export interface OrganizationsRow {
...
+  transactionID: number,
+  /** 
+   * The change in price to the transaction
+   * sum(transaction->entries.amount) = the amount already charged to the account
+   * sum(transactionEntry->ledgerEntries.amount) = transactionEntry.amount
+   */
+  amount: number,
+  /** Description of the entry. Why did the price change */
+  description: string,
+};
```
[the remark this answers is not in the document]
cowsed: I do not  believe so

CurrencyLedger is a ledger of exactly when money changes hands in TB or CC. 
It will never forget a change, it will be a correct list of when we payed or got money

Transactions operate somewhat independently, it keeps track of what net price is, when the price changed, but don't keep track of which accounts it gets routed to.

If a person says hey, why'd you charge this, the Transaction records what it was supposed to do and CurrencyLedger records what happened so you can find where they diverged. 

If you combine the two, you lose out on the what actually happened while making the implementation a lot less intertwined

Transactions are (almost) independent of what actually handles the money, it for the most part just cares that money was handled

server/src/integrations/currency/currency.ts
cowsed: updated comment

Issue #932, opened 2025-09-03
Makerspaces are not clickable enough


JimSHED:
I wish I was joking.

People still do not know they are buttons.

Idk what to do to make it more obvious. But multiple students and professors have said they couldn't tell.

PR #933, opened 2025-09-03
Make makerspace card more clickable

cowsed:
Closes: #932 
Old: 

image

New: 
image


JimSHED:
Let's be as blunt as possible, make it "Learn More" so verbage matches our equipment cards and then I think we are set. Thank you for turning that around so quick!

ZZwatermelon (approved):

PR #950, opened 2025-09-04
Mobile Hours Space Savings

### LINKED ISSUE #926: Mobile Makerspace Hours Formatting
Currently the Makerpsace hours are stacked on mobile, like this:

> Sunday
> 10:00 AM - 2:00 PM

It would be better if they switched to inline ot save vertical space, and lookedl ike this:

> Sunday 10:00 AM - 2:00 PM

Basically they should be a stack in the row direction with justifyContent="space-between"

ZZwatermelon:
Closes #926


cowsed (approved):
Is it really that easy

PR #978, opened 2025-09-09
Allow searching by number and text in CurrencyLedger + Performance improvements

cowsed:
made it so you can search text fields like username (idk1234) or description (job id 45) by number without it assuming you are searching only number fields

Also, rearrange page so that typing doesnt trigger RequestWrapper callback to rerender entire datagrid so typing is a bit snappier


ZZwatermelon (approved):
what if I wanted data-grid to rerender every time I type a character?

cowsed:
> what if I wanted data-grid to rerender every time I type a character?

sorry for ruining your vision

Issue #482, opened 2025-05-07
Test Fail Page Updates


JimSHED:
If you fail a test, the fail page should have a "retake" button, as well as the count of how many times you can try it today with a brief explanation of the training lockout system.

PR #984, opened 2025-09-12
Quiz fail page retake section

yl9431:
Closes: #482

### DIFF
--- a/client/src/pages/maker/take_quiz/EquipmentProgessCard.tsx
+++ b/client/src/pages/maker/take_quiz/EquipmentProgessCard.tsx
@@ -29,7 +29,7 @@ export default function EquipmentProgressCard(props: { moduleID: number }) {
   console.log("should hide", props.moduleID, modulesToHideInPersonFor, shouldHideCompetency)
 
   return (
-    <Card sx={{ width: isMobile ? "100%" : "50%" }}>
+    <Card sx={{ width: (isMobile ? "90vw" : 0.85) }}>
       <CardHeader sx={{ fontWeight: "bold" }} title="Next for you"></CardHeader>
       <RequestWrapper loading={accessProgressResult.loading} error={accessProgressResult.error}>
         <CardContent>
--- a/client/src/pages/maker/take_quiz/RetakeQuiz.tsx
+++ b/client/src/pages/maker/take_quiz/RetakeQuiz.tsx
@@ -18,7 +18,7 @@ export default function RetakeQuiz(props: { moduleID: number }) {
   return (
     <RequestWrapper loading={ submissions.loading || submissions.data === undefined }
       error={ submissions.error } >
-    <Card sx={{ width: isMobile ? "100%" : "50%" }}>
+    <Card sx={{ width: (isMobile ? "90vw" : 0.85) }}>
       <CardHeader sx={{ fontWeight: "bold" }} title="Retake This Quiz"></CardHeader>
       <CardContent>
         <Card>


ZZwatermelon (changes requested):

client/vite-env.d.ts
```diff
@@ -13,6 +13,7 @@ interface ImportMetaEnv {
     readonly VITE_EQUIPMENT_IDS_WITHOUT_INPERSON: string;
     readonly VITE_MODULE_IDS_WITHOUT_INPERSON: string;
     readonly VITE_HELP_PAGE_URL: string;
+    readonly VITE_TRAINING_MAX_ATTEMPTS_PER_DAY_BEFORE_LOCK: string;
```
cowsed, line 16: I like the "you have 4 attempts left warning" but this env variable somewhat duplicated `TRAINING_MAX_ATTEMPTS_PER_DAY_BEFORE_LOCK` on the server. you have to add it here with the VITE prefix for the client to access it, but would you mind changing the uses of it on the backend to use the VITE prefix as well so we don't have to keep 2 different variables in sync

the client cant see server vars, but the server can see client vars when on production

client/src/pages/maker/take_quiz/EquipmentProgessCard.tsx
cowsed, line 1: none of these changes look really intentional? do you mind resetting them to base just so they dont pollute the git history

client/vite-env.d.ts
cowsed: i just realized this is what stephen said and i agree with him more than myself so this env var shouldnt need to exist. 
I think you should stick to the backend as the source of truth and if you want to perserve the 'X tries', 'Y remaining' you should add those to the module query and have the backend fill them in bc the backennd already tracks that

client/src/pages/maker/take_quiz/QuizResults.tsx
```diff
@@ -45,17 +46,21 @@ export default function QuizResults() {
...
-          error={ moduleResult.error }
-        >
+        loading={moduleResult.loading || moduleResult.data === undefined}
+        error={moduleResult.error}
+      >
         <Page title="Quiz Results">
           <Stack direction={"row"} justifyContent={"flex-start"} width={"100%"}>
             <Stack direction="column" width={isMobile ? "100%" : "50%"}>
-              <SubmissionCard module={moduleResult.data?.module!} submission={submissionResult.data?.latestSubmission}/>
+              <SubmissionCard module={moduleResult.data?.module!} submission={submissionResult.data?.latestSubmission} />
```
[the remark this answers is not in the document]
ZZwatermelon: valid

client/src/pages/maker/take_quiz/RetakeQuiz.tsx
```diff
@@ -0,0 +1,42 @@
...
+    );
+
+    return ( 
+      <Card sx={{ width: isMobile ? "100%" : "50%" }}>
+        <CardHeader sx={{ fontWeight: "bold" }} title="Retake This Quiz"></CardHeader>
+        <CardContent>
+          <Card>
+            <CardContent>
+              <Grid container direction={isMobile ? "column" : "row"} >
+                    {Number(submissions.data?.remainingSubmissions.submissions) >= Number(submissions.data?.remainingSubmissions.submissionLimit) ? // TO-DO 9/15: find way to get remaining attempts here
```
ZZwatermelon, line 23: It seems like you achieved this TODO, no? If you have the number of submissions & the limit, can't you just do `limit - submissions`

(As a note I'm not super familiar with the system that allows staff to lift training holds, so you could maybe end up with a negative number or some weird behavior sometimes, but that's not the end of the world IMO)

server/src/resolvers/trainingSubmissionResolver.ts
```diff
@@ -54,7 +54,24 @@ const TrainingSubmissionResolvers = {
...
+     * @throws GraphQLError if not authenticated or is on hold
+     */
+    remainingSubmissions: async (
+      _parent: any,
+      args: { moduleID: string },
+      { ifAuthenticated }: ApolloContext
+    ) => 
+      ifAuthenticated (async (user: any) => {
+        var attempts = (await SubmissionRepo.getFailedSubmissionsTodayByModuleAndUser(Number(args.moduleID), user.id)).length
+        return {"submissions" : attempts, "submissionLimit" : Number(process.env.TRAINING_MAX_ATTEMPTS_PER_DAY_BEFORE_LOCK)};
```
ZZwatermelon, line 72: Maybe make this more verbose and call it failed submissions?

ZZwatermelon (changes requested):

server/src/resolvers/trainingSubmissionResolver.ts
```diff
@@ -54,7 +54,24 @@ const TrainingSubmissionResolvers = {
...
-    })
+    }),
+
+    /**
+     * Fetch failed submissions made by a user for a module
+     * @argument moduleId ID of the TrainingModule 
+     * @returns number of attempts made and limit allowed
+     * @throws GraphQLError if not authenticated or is on hold
+     */
+    remainingSubmissions: async (
```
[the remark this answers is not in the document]
cowsed: https://github.com/rit-construct-makerspace/access-control-server/pull/984#discussion_r2352622147
i think the resolver probably should combine the information (tho not necessarily in this way) bc of the point you make up here

i think a negative showing up for num remaining is quite strange and avoidable so i think this resolver may as well be combined bc both # of attempts taken and # of attempts remaining are dependent on the training id since you could fail 3 times, get the hold removed, fail again in which case it would be 4 taken, 2 remaining whcih wouldn't work if submission limit just returned the env variable

would have to look a little more into how training holds work but i dont think jsut returning the env variable for everything always is a good idea bc we have the ability to lift training holds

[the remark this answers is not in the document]
cowsed: ig we could have them be separate calls but they're very related and if you really want only one, you could just omit it in the gql query from the front end

server/src/schemas/trainingSubmissionSchema.ts
```diff
@@ -26,9 +26,15 @@ export const TrainingSubmissionTypeDefs = gql`
...
+  type remainingSubmissions{
+    submissions : String
+    submissionLimit : String
+  }
+
   extend type Query {
     submission(submissionID: ID): [Submission]
     submissions(moduleID: ID): [Submission]
     latestSubmission(moduleID: ID): Submission
+    remainingSubmissions(moduleID: ID): remainingSubmissions
```
ZZwatermelon, line 38: @cowsed is there an argument that this should take a userID as well so in the future it can be used for staff to see how many attempts a user has on a training? Currently it just works off of whatever user is making the request.

server/src/resolvers/trainingSubmissionResolver.ts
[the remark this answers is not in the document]
yl9431: Kind of relating to the other comments about this, the training holds are created upon module submission so presumably even if they get the hold lifted without giving them extra attempts/deleting previous submissions they'll just create another one immediately.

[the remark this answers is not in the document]
ZZwatermelon: Since it is just one more attempt and they get immediately locked out, then yeah  no reason to worry about negative numbers I 'spose

server/src/schemas/trainingSubmissionSchema.ts
cowsed: yeah i do think this and latestSubmission could be useful in the future
though i think that then pushes me back to the side of submissionsRemaining should also be according to (userID, moduleID) rather than the negative = 0 trick bc a staff would want to know that as well. 

or ig you can negative->0 in both places but that seems a little messy idk

yl9431:
Did you grab the screenshot as the page was refreshing or did it stay like that? I know that appears while the page refreshes but if it remains like that it's a separate issue.

client/src/pages/maker/take_quiz/QuizResults.tsx
```diff
@@ -45,17 +46,21 @@ export default function QuizResults() {
       error={ submissionResult.error }
     >
```
cowsed, line 47: im honestly not really sure what happened here but somehow the next steps card got kind of goofy for mobile? dyk how that happened/how to fix it?

on this branch:
image

Before:
image

server/src/resolvers/trainingSubmissionResolver.ts
cowsed: ^

client/src/pages/maker/take_quiz/QuizResults.tsx
yl9431: Probably just some width issues for the the different components. I'll take a look

ZZwatermelon (changes requested):

client/src/pages/maker/take_quiz/QuizResults.tsx
```diff
@@ -37,27 +38,29 @@ export default function QuizResults() {
           window.removeEventListener('resize', handleWindowSizeChange);
       }
   }, []);
-  const isMobile = width <= 768;
+  const isMobile = useIsMobile();
```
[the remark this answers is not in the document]
yl9431: I made another pr and just did the other issue as well. It should all be cleaned up over there.

cowsed:
image
it seems like when i fail a quiz, it gets stuck loading forever? does that also happen to you
also, it seems like the next for you section got half sized again

Issue #791, opened 2025-08-07
Replace user modal with a dedicated user page


ZZwatermelon:
Since opening the user modal resets the search queries, we don't get a lot of the benefits of using a modal. Upon clicking a user's card, the user should instead be directed to a dedicated page for that user which contains all the content that the user modal does currently.

PR #1008, opened 2025-09-16
dedicated user page

yl9431:
Closes: #791 
Closes: #1025


cowsed:
Could you make the X an 'Back to People' or 'All People' Button. I think its good that it brings you to the users page but the X is a little unclear and just kinda seems leftover from when it was a modal

PR #1020, opened 2025-09-21
PETG Workgroup (+ easier workgroup adding in the middle-term)

cowsed:
Long term, there will be quiz-maker selectable actions for what to do on completion
For now, just made loading them only have you add a number to 2 lists instead of editting server code

introduces:
ID_3DPRINTEROS_QUIZ_WG_MAPPINGS_QUIZZES: string containing comma separated integer module ids
ID_3DPRINTEROS_QUIZ_WG_MAPPINGS_WORKGROUPS: string containing comma separated integer workgroup ids

### DIFF
--- a/server/src/data/humanReadableNames.ts
+++ b/server/src/data/humanReadableNames.ts
@@ -10,7 +10,7 @@ export function generateRandomHumanName(): string {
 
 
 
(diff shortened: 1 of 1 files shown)


cowsed:
wow great work me

Issue #1019, opened 2025-09-21
Expanding Mobile Staff Dropdown then resizing window to be full size again compresses Staff bar strangely


cowsed:
Doesnt happen if the menu is closed before resizing

Before:

<img width="1920" height="894" alt="Image" src="https://github.com/user-attachments/assets/0fc0b326-54cf-42b0-839d-5e0b198c8065" />

During: 

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/478f2f57-a33b-45eb-9b49-5f2b790cbacf" />

After:

<img width="1920" height="883" alt="Image" src="https://github.com/user-attachments/assets/39c2b296-3735-4bf4-8094-320356b66330" />

PR #1055, opened 2025-09-26
Responsive (Mobile) Staff Bar 

SierraTran:
Closes: https://github.com/rit-construct-makerspace/access-control-server/issues/1019

### DIFF
--- a/client/src/pages/makerspace_page/StaffBar.tsx
+++ b/client/src/pages/makerspace_page/StaffBar.tsx
@@ -1,23 +1,23 @@
-import { ButtonBase, Stack, Typography } from "@mui/material";
-import { useCurrentUser } from "../../common/CurrentUserProvider";
-import NavLink from "../../top_nav/NavLink";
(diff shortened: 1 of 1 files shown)


client/src/pages/makerspace_page/StaffBar.tsx
```diff
@@ -42,7 +42,7 @@ export default function StaffBar() {
         padding="10px 0px"
       >
         {
-          mobileMenu
+          isMobile && mobileMenu
```
SierraTran, line 45: I just added to the conditional statements, so there's no more useEffect.

client/src/pages/makerspace_page/StaffBar.tsx
```diff
@@ -75,7 +75,7 @@ export default function StaffBar() {
...
```
SierraTran, line 78: I also added it here, too.

cowsed (approved):
Looks good

ZZwatermelon (approved):

Issue #546, opened 2025-05-21
Page on Passed Trainings


JimSHED:
When you click on a training you passed, it should give a page with info on your passing score, when you passed, etc., and a notice of when it will expire. Then an option to re-take it, versus the current system where it just re-opens the quiz. 

Maybe we also show the graded result from when they took the quiz? That may make it too easy to share answers though.

PR #1061, opened 2025-09-26
passed trainings results page

yl9431:
Closes: #546


yl9431:
Ok I'll leave it as retake if they failed and renew if they've passed.

cowsed:
This looks good, will rereview when it merges main just to make sure its all still good since this is awhile behind main

Issue #1037, opened 2025-09-25
Use url params for search query to enable linking to most everythinh


cowsed:
on some pages, when you type in the search box, the query is stored as a url param (finances page is a good example of this).
This will let us link to helpful things in
- Makerspace page (equipment search)
- readers page
- probably othres but i dont remember them off the top of my head

yl9431:
For the other pages do we still want the url to only be updated upon submitting or do we want to update it dynamically like for equipment search

cowsed:
i think you can make them update as the page content updates. if the page updates every keystroke, so should the url. If it waits for an enter, the url should also wait for the error

PR #1078, opened 2025-10-01
url params for search queries

yl9431:
Closes: #1037

### DIFF
--- a/client/src/pages/lab_management/inventory/InventoryPage.tsx
+++ b/client/src/pages/lab_management/inventory/InventoryPage.tsx
@@ -1,4 +1,4 @@
-import { useState } from "react";
+import { useEffect, useState } from "react";
 import { Box, Button, Stack, Typography } from "@mui/material";
 import SearchBar from "../../../common/SearchBar";
 import CreateIcon from "@mui/icons-material/Create";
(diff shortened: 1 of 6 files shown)


yl9431:
Might've missed some search bars, let me know if you find one.
This seems to bring back the issue of "delayed inputs" if you type really fast (especially for the first few letters), since the page re-renders whenever you type something.

cowsed:
I wonder if debouncing the search bar could help with the delayed inputs, so it only updates if you havent typed for 100ms or whatever

client/src/pages/lab_management/inventory/InventoryPage.tsx
```diff
@@ -19,10 +20,24 @@ export default function InventoryPage() {
...
+
   const inventoryTagsResult = useQuery(GET_INVENTORY_TAGS);
 
   const makerspacesWithItemsResult = useQuery(GET_MAKERSPACES_WITH_ITEMS);
 
+  const setUrlParam = (paramName: string, paramValue: string) => {
+    const params = new URLSearchParams(location.search);
+    params.set(paramName, paramValue);
+    navigate(`/admin/inventory?` + params, { replace: true });
+  };
```
ZZwatermelon, line 33: I mean here instead of navigating to it on every change, try only navigation to it onSubmit?

There might be implications with that, not sure tho

yl9431:
That would mean the visible url would only update after hitting enter in the search bar, and we probably wouldn't even need to call seturlParam in onChange. Also the difference in how the page re-renders as you type vs. url only changing on hitting enter.

cowsed:
Yeah the inventory page is doing something silly (loading like 30 4k images at once) so it will just be slow

Issue #1059, opened 2025-09-26
Statistics should report makerspace tap-ins, not room tap-ins


JimSHED:
"Today's Numbers" currently shows tap-in of all rooms, not makerspaces. This means when you tap into the Atrium, it 4x counts it. Throws off at-a-glance numbers.

ZZwatermelon:
@cowsed I think we should just actually remove tapping into a room, and just make the rooms check if you've tapped into their associated makerspace

cowsed:
true, at one point there was a branch that did this but I think it got lost in pursuit of bigger and better things after I got discouraged by having to redo the stats page. 
It took all Room swipes, consolidated multiple into 1 if they were the same tap and dropped it into MakerspaceSwipes and from there all the tapping goes to there

ZZwatermelon:
Yeah I think atp it's just something that needs to happen

The stats page already needs to be reworked on account of it crashing the website

JimSHED:
If the answer to re-working the stats page is Grafana, I am OK with that as well. We'll set it up properly on Heroku.

PR #1081, opened 2025-10-02
Room Deletion Attempt 2

SierraTran:
Closes: #

Redoing my room deletion and room archiving/unarchiving with the expandable header.

### DIFF
--- a/client/src/pages/makerspace_page/MonitorRoomPage.tsx
+++ b/client/src/pages/makerspace_page/MonitorRoomPage.tsx
@@ -1,28 +1,27 @@
 import { useState } from "react";
-import {
-  Box,
-  Button,
-  Stack,
-  TextField,
-  Typography,
-} from "@mui/material";
+import { Box, Button, Stack, TextField, Typography } from "@mui/material";
 import { gql, useMutation, useQuery } from "@apollo/client";
 import { useNavigate, useParams } from "react-router-dom";
 import RequestWrapper2 from "../../common/RequestWrapper2";
 import RoomMakerspaceAssociation from "./RoomMakerspaceAssociation";
-import { DELETE_ROOM, UPDATE_ROOM_NAME } from "../../queries/roomQueries";
-import DeleteIcon from '@mui/icons-material/Delete';
+import { ARCHIVE_ROOM, DELETE_ROOM, UNARCHIVE_ROOM, UPDATE_ROOM_NAME } from "../../queries/roomQueries";
+import ArchiveIcon from "@mui/icons-material/Archive";
+import UnarchiveIcon from "@mui/icons-material/Unarchive";
+import DeleteIcon from "@mui/icons-material/Delete";
 import { useCurrentUser } from "../../common/CurrentUserProvider";
-import SaveIcon from '@mui/icons-material/Save';
+import SaveIcon from "@mui/icons-material/Save";
 import { isManagerFor } from "../../common/PrivilegeUtils";
 import { useIsMobile } from "../../common/IsMobileProvider";
 import { TrainingModule } from "../../common/TrainingModuleUtils";
 import ManageRoomTrainings from "./ManageRoomTrainings";
+import { toast } from "react-toastify";
+import EquipmentBlockDialog from "./EquipmentBlockDialog";
 
 export const GET_ROOM = gql`
   query GetRoom($id: ID!) {
     room(id: $id) {
       name
+      archived
       makerspace {
         id
         name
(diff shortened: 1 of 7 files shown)


cowsed:
I like the popup for equipment blocking it. 
Thought it looks like the room swipes still blocks it (if anyone has ever swiped into it)
image

ZZwatermelon:
> I like the popup for equipment blocking it. Thought it looks like the room swipes still blocks it (if anyone has ever swiped into it) 

I mean how do we work around that? As part of #1059 we could store the swipe as (makerspaceID, userID, date, makerspaceName) then onUpdate null the makerspaceID and present statistics based on makerspaceName. Thoughts @cowsed?

cowsed:
I was under the impression that this was the implementation that recognized that ***everything*** associated with tbat room is purged forever and that you recognize the risk in that. 

If you accidentally make a room you have no swipes so it's not an issue
If you really want to purge this room from history, this lets you do that

@JimSHED can you confirm or deny

ZZwatermelon:
From my perspective though even if you want to explode everything related to that room, you probably want to keep the record of how many people were using that room

JimSHED:
In my mind this was a "oh crap I hit the new room button, lemme undo that."

We will very rarely, if ever, properly delete a room without also "deleting" (archiving) the makerspace. And when we do, we'd probably start with a full export of all data relevant to the room.

client/src/pages/makerspace_page/MonitorRoomPage.tsx
```diff
@@ -76,30 +75,70 @@ export default function ManageRoomPage() {
...
+    try {
       await deleteRoom({
-        variables: { id: roomID }
-      })
-      navigate(`/makerspace/${makerspaceID}/edit`)
+        variables: { id: roomID },
+      });
+      navigate(`/makerspace/${makerspaceID}/edit`);
+    } catch (error: any) {
+      console.log(error);
```
ZZwatermelon, line 135: Don't want to keep the console.log in

client/src/pages/makerspace_page/MonitorRoomPage.tsx
```diff
@@ -120,29 +158,62 @@ export default function ManageRoomPage() {
...
+                      Archive Room
+                    </Button>
+                  )}
+                  {isManagerFor(user, Number(roomID)) ? (
+                    <Button color="error" variant="contained" startIcon={<DeleteIcon />} onClick={handleDeleteRoom}>
                       Delete Room
                     </Button>
-                    : null
-                }
+                  ) : null}
```
ZZwatermelon, line 189: Instead of hiding the button based on managerness, we could just disable the button, and I personally like that more because it keeps the visual experience similar between users

client/src/pages/makerspace_page/MonitorRoomPage.tsx
```diff
@@ -120,29 +158,62 @@ export default function ManageRoomPage() {
...
+              >
+                <Typography variant={"h4"}>
+                  Manage {room.name} [ID: {roomID}]
+                </Typography>
+                <Stack direction={isMobile ? "column" : "row"} spacing={2}>
+                  {room.archived ? (
+                    <Button
+                      color="primary"
+                      variant="contained"
+                      startIcon={<UnarchiveIcon />}
```
ZZwatermelon, line 175: Historically we've used the PublishIcon as opposed to the UnarchiveIcon, which I think I agree with because the archive / unarchive icons look pretty similar to me. I'd welcome other people's opinion on that, but we should be consistent.

client/src/pages/makerspace_page/MonitorRoomPage.tsx
```diff
@@ -120,29 +158,62 @@ export default function ManageRoomPage() {
...
+                alignItems="flex-end"
+                spacing={2}
+              >
+                <Typography variant={"h4"}>
+                  Manage {room.name} [ID: {roomID}]
+                </Typography>
+                <Stack direction={isMobile ? "column" : "row"} spacing={2}>
+                  {room.archived ? (
+                    <Button
+                      color="primary"
```
ZZwatermelon, line 173: We usually use `color="success"` for publishing

server/src/resolvers/roomsResolver.ts
```diff
@@ -85,16 +82,20 @@ const RoomResolvers = {
       }),
 
     archiveRoom: async (_parent: any, args: any) => {
-      return await RoomRepo.archiveRoom(args.id);
+      return await RoomRepo.archiveRoom(args.roomID);
+    },
+
+    unarchiveRoom: async (_parent: any, args: any) => {
+      return await RoomRepo.unarchiveRoom(args.roomID);
```
ZZwatermelon, line 89: These should be restricted using the isManagerFor permission level.

You might need to use the isManager level, then manually check that the user is a manager for the appropriate makerspaces, look how the deleteRoom resolver checks it.

ZZwatermelon (changes requested):

ZZwatermelon (changes requested):
Archive/unarchive room resolvers still need to be protected at the right privilege level

ZZwatermelon (approved):

client/src/pages/makerspace_page/RoomCard.tsx
```diff
@@ -11,13 +12,20 @@ export default function RoomCard(props: RoomCardProps) {
...
+            direction={"row"}
+            spacing={2}
+            sx={{ display: "flex", justifyContent: "space-between", alignItems: "center" }}
+          >
+            <Stack direction={"row"} alignItems={"center"} spacing={1}>
+              {props.room.archived ? <ArchiveIcon sx={{ color: "gray" }} /> : null}
+              <Typography variant="h6" component="div">
+                {props.room.name}
+              </Typography>
+            </Stack>
```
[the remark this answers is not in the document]
cowsed: background being red is kinda garish (see hidden equipment) but it does make it very obvious which i think is good
maybe we could do what the equipment does with the toggle at the top to show archived and when shown, if its archived its red

Issue #1036, opened 2025-09-24
Reformat "Reservation only"


JimSHED:
When you mark a machine as reservation only, it replaces the # of available machines with a message. This looked better on the old card style, but the current card style looks weird. Maybe format it to instead be across the card like the description, or the training info? 

Or just remove the text altogether, and I add it to the description of the equipment. 

<img width="1252" height="398" alt="Image" src="https://github.com/user-attachments/assets/29124dce-7749-4605-bd13-e478d5e62eb2" />

PR #1082, opened 2025-10-02
Move Reservation Only line to below the description

SierraTran:
Closes: #1036


ZZwatermelon (approved):

cowsed:
Looks good, the reason some don't have Number available is because some machines have card readers associated with them

cowsed (approved):

Issue #774, opened 2025-08-05
Add ability to delete training modules


cowsed:
> I just mark them as to delete and then reuse them for future trainings
> - jimbo

PR #1101, opened 2025-10-07
Sierra tran/delete training modules

SierraTran:
Closes: #774


SierraTran:
As of right now, I only have the icon-only button for deleting trainings. 

Should I also add a delete button for editing existing trainings and a cancel button for editing new trainings?

Issue #967, opened 2025-09-08
Search for Equipment


JimSHED:
It would be interesting if you could search for equipment, especially now as we start to have more equipment segregated across multiple makerspaces. 

So if you are looking for a laser cutter, you can search that and see where to find them.

PR #1116, opened 2025-10-15
global search

yl9431:
Closes: #967


yl9431:
Currently search results page has equipment, spaces and trainings. Any ideas on what a room card should look like or what kind of information it should contain since rooms are just a location with a name.
Alternatively can scrap using the pre-existing cards and just sort by spaces with their rooms as sections that contain simplified equipment and trainings and just linking to each item kind of like how trainings are currently.

Clicking on a specific item in the autocomplete should link to said item on its corresponding page, which is waiting on #1037.

yl9431:
Also, should the search show archived/hidden items if the user is staff? Or should they be hidden regardless of who is logged in.

Issue #918, opened 2025-08-29
CDN Image Component/Function


cowsed:
Everytime we display an image we have to concatenate the CDN identifier and the CDN URL and fill in the default image if it doesnt have an image.

```tsx
       <CardMedia
          component="img"
          height={props.isMobile ? "197px" : "281px"}
          image={import.meta.env.VITE_CDN_URL + "user-uploads/" + props.imageUrl}
        />
 ```

It would be lovely if we could say
```tsx
<CDNImage width={} height={} identifier={zone.image} />
```
and it looks up the image or fills in the default.

ZZwatermelon:
Could also just allow the component to take a boolean CardMedia prop that defaults to false.

PR #1117, opened 2025-10-15
function to fix image link or use default

yl9431:
Closes: #918

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -39,7 +39,7 @@
         "typescript": "^5",
         "use-immer": "^0.11.0",
         "uuid": "^8.3.2",
-        "vite": "^7.1.5",
+        "vite": "^7.1.11",
         "vite-plugin-eslint2": "^5.0.4",
         "web-vitals": "^2.1.4"
       },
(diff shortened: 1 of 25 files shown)


client/src/common/ImageFinder.tsx
ZZwatermelon: When implementing this you should go through the codebase and make everywhere that uses an image on AWS use this

client/src/common/ImageFinder.tsx
```diff
@@ -0,0 +1,4 @@
+export function findImage(image: string|undefined|null, path= "", ){
```
ZZwatermelon: why's there another comma after path? Is there gonna be a 3rd argument?

ZZwatermelon (dismissed):

client/src/common/ImageFinder.tsx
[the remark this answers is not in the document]
cowsed: yeah, makeCDNLink seems good. I would also make sure to add a comment about what the path does and what the image arg should be (date image name)

PR #1124, opened 2025-10-20
fix module resolution 

cowsed:
(shouldve been bundler bc using ts ^5  allows for moduleResolution bundler)


ZZwatermelon (approved):
yippee 🇬🇺

Issue #1062, opened 2025-09-26
Announcements Optional Link button


JimSHED:
We sometimes need to make an announcement with a call to action to another page, and/or a longer announcement that could instead be on the docs and we just need to point to it.

There should be an option when writing an announcement to have a button, like the "Learn More" on equipment, that we can customize the text on and then have a link that opens in new tab. 

Examples of the 3 current ones;

<img width="1267" height="466" alt="Image" src="https://github.com/user-attachments/assets/7a487a90-b0ae-4a30-a753-690e4f9c4324" />

* Welding one would be 2 sentences, then a link at the bottom "Click Here to Learn More!"
* PETG would have "Click Here to Take the Quiz!"
* Maker Faire would have "Learn more at rochester.makerfaire.com"

Issue #1054, opened 2025-09-26
Give a preview of the announcement cards when editing or creating an announcement


ZZwatermelon:
Similar to how we give a preview of the equipment card & makerspace card when editing those

PR #1139, opened 2025-10-24
Announcement Preview and Link Button

SierraTran:
Closes: #1062 
Closes: #1054

### DIFF
--- a/client/src/common/CurrentUserProvider.tsx
+++ b/client/src/common/CurrentUserProvider.tsx
@@ -1,46 +1,8 @@
 import { createContext, ReactElement, useContext } from "react";
-import { gql, useQuery } from "@apollo/client";
+import { useQuery } from "@apollo/client";
 import RequestWrapper2 from "./RequestWrapper2";
 import { Navigate, useLocation } from "react-router-dom";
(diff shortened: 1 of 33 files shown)


ZZwatermelon (dismissed):

client/src/pages/lab_management/announcements/AnnouncementModalContents.tsx
```diff
@@ -32,44 +31,73 @@ export default function AnnouncementModalContents({
   onDelete,
   loading,
 }: AnnouncementPageProps) {
-
   const navigate = useNavigate();
-  
+
+  const [showLinkFields, setShowLinkFields] = useState(!!announcementDraft.linkUrl || !!announcementDraft.linkText);
```
cowsed: When testing this, it seems like the default value of this isnt getting calculated just right. If i add a link and link text to an announcement, then save and close it, then reopen it, the Link Button switch is set to off. If i turn it on manually it shows the saved fields just fine though

cowsed (commented):
I like the default to https, thats really nice. just the one thing about the switch and i think this is good

Issue #1110, opened 2025-10-12
Don't say trainings incomplete if not logged in


JimSHED:
When you are not logged into make.rit.edu, you are treated as if you have no trainings done. This means there are a ton of warnings that you can't really resolve if you don't have an account, and makes it appear that you "lost" your training progress if you do have an account. Some proposed fixes;

* Hide the "All makerspace users must complete the Shop Safety..." banner if not logged in.
* Change "You have incomplete makerspace trainings" to a more neutral color (purple?) that says "Log in to view training progress." instead.
  * Same with "You must pass the makerspace trainings before you can use equipment in the makerspace!" and "You must pass the area trainings before you can use equipment in the area!"
* Change red "X" next to incomplete (all) trainings to something more neutral, match the color to the color of the admonitions above. 
  * My initial; thought was a "?" in purple (matches neutral color and login button), but idk.

JimSHED:
Yeah actually I never thought about the annoyance of not being logged in and getting sent to a completely different page

PR #1142, opened 2025-10-24
Neutral Training Icons for Not-Logged-In Users

SierraTran:
Closes: #1110
Closes: #1146


cowsed:
Kinda scope-creeping this PR, but while you're here, could you change the icon for when the user is logged in but haven't taken the makerspace trainings as in #1146

Issue #1123, opened 2025-10-20
Changes May Not be saved dialog on the equipment editor page


cowsed:
now that theres a lot more stuff on the edit equipment page and a live updating preview, its pretty easy to forget that you edited an option but didnt save it. A thing that warns you when you try to leave the page could be nice here

Issue #1090, opened 2025-10-03
Trainer Route is Unused


ZZwatermelon:
Currently the fronted protects staff-only routes in a variety of ways.

One of these is to wrap the protected pages at the router level to redirect those without proper access.

One of these wrappers is the trainer route, which is currently unused, but should be wrapping the user's page to allow trainers to approve access checks for the machines they are trainers for.

PR #1145, opened 2025-10-28
"Changes may not be saved" Dialog for Equipment Editor Page

SierraTran:
Closes: #1123
Closes: #1090

### DIFF
--- a/client/src/pages/makerspace_page/equipment_pages/EquipmentInformation.tsx
+++ b/client/src/pages/makerspace_page/equipment_pages/EquipmentInformation.tsx
@@ -100,6 +100,8 @@ export default function EquipmentInformation(props: EquipmentInformationProps) {
         requiresInPerson: requiresInPerson,
       },
     });
+
+    setUnsaved(false);
   }
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


client/src/pages/makerspace_page/equipment_pages/EquipmentInformation.tsx
```diff
@@ -88,6 +88,24 @@ export default function EquipmentInformation(props: EquipmentInformationProps) {
...
+      requiresInPerson !== props.equipment.requiresInPerson ||
+      room.id !== props.equipment.room.id ||
+      moduleIDs.length !== props.equipment.trainingModules.length;
+
+    if (unsaved) {
+      event.preventDefault();
+      return '';
+    }
+  });
+
```
cowsed: not a you issue but i just checked on line 153 and you can delete the comment block and ts-ignore now, seeminlgy the api has changed since it was written

cowsed:
Hmmm, from what i can see it looks like createBrowserRouter is the new recommended way of doing things and gives us more features like blocking pages and more. 
Theres a couple places that should have useBlocker like behavior that currently don't behave correctly so i think doing the transition to createBrowserRouter makes sense

we just have to make sure we test all the pages especially ones with more complicated url formats

cowsed:
:) this is pretty great, love the dialog box and it like actually prevents me from leaving the page. 
just one thing, if i make an edit, press the save button, then try to leave the page, it still blocks me though not always with the pop up - sometimes it just does nothing. At this point i think it should let me off the page since i pressed save and the value is set. if i ignore the warning, leave the page and come back, the value is what i set and saved it to, so maybe the check on the front end is just a little over cautious?

Issue #1153, opened 2025-11-03
"Publish" training clips with makerspace dropdown


JimSHED:
<img width="2365" height="250" alt="Image" src="https://github.com/user-attachments/assets/264f0923-8e3b-4dc0-bdc9-8f4131545c3d" />

PR #1166, opened 2025-11-06
Stop "Publish" Button from Clipping into Makerspace Dropdown

SierraTran:
Closes: #1153

Moved the width style from Select to FormControl since FormControl is the parent container. This will tell it to stretch or shrink accordingly.

### DIFF
--- a/client/src/common/CurrentUserProvider.tsx
+++ b/client/src/common/CurrentUserProvider.tsx
@@ -1,46 +1,8 @@
 import { createContext, ReactElement, useContext } from "react";
-import { gql, useQuery } from "@apollo/client";
+import { useQuery } from "@apollo/client";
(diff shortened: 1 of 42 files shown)


client/src/pages/lab_management/edit_module/EditModulePage.tsx
```diff
@@ -130,7 +130,7 @@ export default function EditModulePage({
       <title>Edit Training | Make @ RIT</title>
       <Typography variant="h4" textAlign="center">Edit {module.name}</Typography>
       <Stack
-        direction="row"
+        direction={isMobile ? "column" : "row"}
```
cowsed: this page crashes when i try to load it complaining that isMobile is never defined. I think you need to import useIsMobile and create the isMobile constant in this file as well

Issue #1157, opened 2025-11-04
Trainer: Hide Hold Reasons


JimSHED:
Trainers can/should be able to see if an account has a hold on it on the main people page, but they should not see the specific notes under the "Account Holds" section for current or past holds.

Issue #1156, opened 2025-11-04
Trainer: Disable Buttons


JimSHED:
Trainer roles can still click some buttons on the People page, but it doesn't appear they do anything. Specifically, they have the buttons to update ID card, place hold, and make someone a staff/manager/trainer. 

These should be greyed out.

PR #1171, opened 2025-11-13
Trainer Restrictions for User page

SierraTran:
Hey, @yl9431 ! I saw your pr for the user page revamp and wanted to contribute some code for some trainer restrictions since you're already there. The specific issues I targeted are #1156 and #1157. 

It won't show up without the Trainer Route being used, but I have it used in my other pr (#1145). Hopefully, that one will be merged soon.

I hope this helps!


yl9431 (approved):
Looks good!
I'll update the pr going into main so the issues being addressed here are #1156 and #1157 right?

PR #1341, opened 2026-03-09
WSACS2

ZZwatermelon:


server/src/resolvers/equipmentResolver.ts
```diff
@@ -156,7 +156,7 @@ const EquipmentResolvers = {
 
         const equipment = await EquipmentRepo.addEquipment(args.equipment);
 
-        await createLog(
+        await createUnassocaitedAuditLog(
```
cowsed: should also probably be makerspace log in the future

server/src/auth.ts
```diff
@@ -260,6 +260,7 @@ export function setupDevAuth(app: express.Application) {
         await createLog(
           `{user} logged in.`,
           "server",
+          // @ts-expect-error
```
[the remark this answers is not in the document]
ZZwatermelon: so I mean this is clearly never happening, should I just get rid of these logging portions since it doesn't make sense to do anyway?

server/src/models/devices/accessController.ts
```diff
@@ -1,15 +1,144 @@
...
     this.channelID = rawRow.channelID;
     this.state = rawRow.state;
+    this.tempDuration = rawRow.tempDuration;
+  }
+
+  async canUnlock(userID: number, log?: boolean): Promise<{ hasAccess: boolean, reason: AccessAttemptReason }> {
+    const rawUser = await UserRepo.getUserByIDOrUndefined(userID);
+    if (rawUser === undefined) {
+      if (log) {
+        UnlockAttemptRepo.createUnlockAttemptLog(undefined, "NO_USER", undefined, "UNKOWN_USER", false, AccessAttemptReason.UNKNOWN_USER);
```
cowsed: for the first NO_USER shouldnt that be "none" or "" or something because the second arg is the equipment

likewise at line 65 i think  it should be none or empty since UNPAIRED is not the name of the equipment

server/src/models/devices/accessController.ts
```diff
@@ -1,15 +1,144 @@
...
+          { id: this.deviceID, label: device?.name ?? "UNKNOWN DEVICE" }
+        );
+      }
+      return { hasAccess: false, reason: AccessAttemptReason.UNPAIRED };
+    }
+
+    const rawEquipment = await EquipmentRepo.getEquipmentOrUndefinedByID(instance.id);
+    if (rawEquipment === undefined) {
+      if (log) {
+        UnlockAttemptRepo.createUnlockAttemptLog(undefined, "UNPAIRED", user.id, user.ritUsername, false, AccessAttemptReason.UNPAIRED);
```
cowsed: see comment above about none rather than UNPAIRED

also should we have different error reasons for unpaired to access controller vs unpaired to equipment?  i kinda think no but it could be helpful for debugging things

server/src/models/api/WSACSController.ts
```diff
@@ -0,0 +1,311 @@
...
+    });
+
+    ws.onclose = (event) => this.handleWsClose(event, deviceID);
+    ws.onerror = (event) => this.handleWsError(event, deviceID);
+    ws.onmessage = (event) => this.handleWsMessage(event, deviceID);
+  }
+
+  static sendCoreRequest(payload: WSACSServerUnprompted, deviceID: number) {
+    const connection = this.corePool.get(deviceID);
+    if (connection === undefined) { return; }
```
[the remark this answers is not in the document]
cowsed: same with sendCoreResponse

PR #1419, opened 2026-05-12
Always regenerate atrium tokens on log in

cowsed:


ZZwatermelon (approved):
Did you mean to leave the log statement?

cowsed:
> Did you mean to leave the log statement?

yes for now at least bc i wanna see when it happens since i cant test w/ real shib

